PR #9, opened 2021-09-02
support moving ctdb enabled pods to new addresses

phlogistonjohn:
The pod that acts as a CTDB node may be migrated/restarted/or otherwise change IP address. Tweaking our approach to populating the nodes file we can update the CTDB cluster to use the node with it's new address.

Please review by 2021-09-09


spuiuk:
Still pending on me. Please give me another couple of days.

spuiuk (approved):
Sorry for the time taken in going through the patches. I have gone through the set and they look good to me.

phlogistonjohn:
> Sorry for the time taken in going through the patches. I have gone through the set and they look good to me.

Thanks! I do appreciate it. :-)

PR #14, opened 2021-09-21
Git checkout to a branch to avoid detached HEAD

anoopcs9:
Minor improvement to hide the following advice:

_Cloning into '/var/tmp/build/sambacc'...
Note: switching to 'e3e2a53bea5d'. You are in 'detached HEAD' state._


phlogistonjohn (approved):
Seems OK, thanks :-)

PR #17, opened 2021-09-30
Adapt to samba features/changes using env variable `SAMBA_SPECIFICS`

### LINKED ISSUE #15: Support conditional behavior for differing samba versions
Specifically samba 4.15 has made changes to some of the command line options. As noted in [a samba-container issue](https://github.com/samba-in-kubernetes/samba-container/issues/45) the option `--log-stdout` has been changed to `--debug-output`.

I'm proposing we add a mechanism for handling small behavior changes across samba versions like this.

Currently I'm thinking we can add an environment variable that indicates the various behaviors. I'm thinking something like `SAMBA_FEATURES` and our initial value can be `SAMBA_FEATURES=cli_debug_output`. In the future other behaviors could be tracked like `SAMBA_FEATURES=cli_debug_output,foo_bar,something_cool`.

anoopcs9:
Fixes #15

### DIFF
--- a/sambacc/samba_cmds.py
+++ b/sambacc/samba_cmds.py
@@ -28,12 +28,24 @@ _GLOBAL_PREFIX: typing.List[str] = []
 _GLOBAL_DEBUG: str = ""
 
 
+# Known flags for SAMBA_SPECIFICS env variable
+_SMBD_CLI_STDOUT_OPT: str = "smbd_cli_debug_output"
+
+
 def get_samba_specifics() -> typing.Set[str]:
     value = os.environ.get("SAMBA_SPECIFICS", "")
     out = set()
     for v in value.split(","):
         out.add(v)
     return out
+
+
+def _smbd_stdout_opt() -> str:
+    opt = "--log-stdout"
+    opt_lst = get_samba_specifics()
+    if _SMBD_CLI_STDOUT_OPT in opt_lst:
+        opt = "--debug-stdout"
+    return opt
 
 
 def set_global_prefix(lst: typing.List[str]) -> None:
@@ -97,11 +109,10 @@ smbd = SambaCommand("/usr/sbin/smbd")
 
 winbindd = SambaCommand("/usr/sbin/winbindd")
 
-smbd_foreground = smbd[
-    "--foreground",
-    "--log-stdout",
-    "--no-process-group",
-]
+
+def smbd_foreground():
+    return smbd["--foreground", _smbd_stdout_opt(), "--no-process-group"]
+
 
 winbindd_foreground = winbindd[
     "--foreground",
--- a/sambacc/samba_cmds.py
+++ b/sambacc/samba_cmds.py
@@ -28,6 +28,14 @@ _GLOBAL_PREFIX: typing.List[str] = []
 _GLOBAL_DEBUG: str = ""
 
 
+def get_samba_specifics() -> typing.Set[str]:
+    value = os.environ.get("SAMBA_SPECIFICS", "")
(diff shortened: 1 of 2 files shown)


sambacc/samba_cmds.py
```diff
@@ -28,6 +28,14 @@
 _GLOBAL_DEBUG: str = ""
 
 
+def get_samba_specifics() -> list:
```
phlogistonjohn, line 31: Let's return a set here. Most of the time we'd be wanting to check for the presence of a specific entry and I think that fits the set's semantics better. We should also use something like `typing.Set[str]` for our return since we know the members will be strings.

sambacc/samba_cmds.py
```diff
@@ -36,6 +36,14 @@ def get_samba_specifics() -> list:
     return out
 
 
+def smbd_stdout_opt() -> str:
+    opt = "--log-stdout"
+    opt_lst = get_samba_specifics()
+    if "smbd_cli_debug_output" in opt_lst:
```
phlogistonjohn, line 42: Let's make all the entries we can look for in `SAMBA_SPECIFICS` top level variables so that its easy for future readers of the code to find all the possible values that we look for.

sambacc/samba_cmds.py
```diff
@@ -99,7 +107,7 @@ def __repr__(self):
 
 smbd_foreground = smbd[
     "--foreground",
-    "--log-stdout",
+    smbd_stdout_opt(),
```
phlogistonjohn, line 110: This causes the environment variable to be checked at import time rather than when `smbd_foreground` is used. I'd rather switch smbd_foreground to a function so that the state of the module as imported is not dependent on the environment. This does mean changing the users of `smbd_foreground` but I think it's a small price to pay.

sambacc/samba_cmds.py
```diff
@@ -36,6 +36,14 @@ def get_samba_specifics() -> list:
     return out
 
 
+def smbd_stdout_opt() -> str:
```
phlogistonjohn, line 39: I doubt we'll need this function outside of samba_cmds.py let's prefix it with an underscore (_smbd_stdout_opt) to make it "private".

phlogistonjohn (changes requested):
Thanks a bunch for this. I've got a few remarks that I think will make the code more maintainable long term, but overall a really good start thanks!

sambacc/samba_cmds.py
anoopcs9: Done.

sambacc/samba_cmds.py
anoopcs9: Yes. Good thought..

sambacc/samba_cmds.py
anoopcs9: I am updating the patch set with my python knowledge. Feel free to comment if I have to change something more.

sambacc/samba_cmds.py
```diff
@@ -28,6 +28,10 @@
 _GLOBAL_DEBUG: str = ""
 
 
+"""Recognizable strings for SAMBA_SPECIFICS env variable"""
```
phlogistonjohn, line 31: This should probably be just a comment:
```suggestion
# Known flags for SAMBA_SPECIFICS env variable
```

phlogistonjohn (changes requested):
Just one more small thing but we're nearly there.

sambacc/samba_cmds.py
anoopcs9: Done

PR #18, opened 2021-10-05
Add support for new/replaced winbindd cli option `--debug-stdout`

anoopcs9:
See sub-section **_Duplicates in command line utils_** under _Improved command line user experience_ from [v4.15 release notes](https://www.samba.org/samba/history/samba-4.15.0.html)

### DIFF
--- a/sambacc/samba_cmds.py
+++ b/sambacc/samba_cmds.py
@@ -29,7 +29,7 @@ _GLOBAL_DEBUG: str = ""
 
 
 # Known flags for SAMBA_SPECIFICS env variable
-_SMBD_CLI_STDOUT_OPT: str = "smbd_cli_debug_output"
+_DAEMON_CLI_STDOUT_OPT: str = "daemon_cli_debug_output"
 
 
 def get_samba_specifics() -> typing.Set[str]:
@@ -40,10 +40,13 @@ def get_samba_specifics() -> typing.Set[str]:
     return out
 
 
-def _smbd_stdout_opt() -> str:
-    opt = "--log-stdout"
+def _daemon_stdout_opt(daemon: str) -> str:
+    if daemon == "smbd":
+        opt = "--log-stdout"
+    else:
+        opt = "--stdout"
     opt_lst = get_samba_specifics()
-    if _SMBD_CLI_STDOUT_OPT in opt_lst:
+    if _DAEMON_CLI_STDOUT_OPT in opt_lst:
         opt = "--debug-stdout"
     return opt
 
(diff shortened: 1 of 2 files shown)


sambacc/samba_cmds.py
```diff
@@ -30,6 +30,7 @@
 
 # Known flags for SAMBA_SPECIFICS env variable
 _SMBD_CLI_STDOUT_OPT: str = "smbd_cli_debug_output"
+_WINBINDD_CLI_STDOUT_OPT: str = "winbindd_cli_debug_output"
```
phlogistonjohn, line 33: While I generally like granular options, is it ever possible that winbindd and smbd would have different CLI options? In other words, perhaps it would be better to just combine these into one?
This isn't a demand though, I'd like to hear your thoughts.

anoopcs9: With new/improved _command line user experience_, options have become more organized. We now have some options common to different daemons listed as follows:

```
Help options:
  -?, --help                         Show this help message
      --usage                        Display brief usage message

Common Samba options:
  -d, --debuglevel=DEBUGLEVEL        Set debug level
      --debug-stdout                 Send debug output to standard output
  -s, --configfile=CONFIGFILE        Use alternative configuration file
      --option=name=value            Set smb.conf option from command line
  -l, --log-basename=LOGFILEBASE     Basename for log/debug files
      --leak-report                  enable talloc leak reporting on exit
      --leak-report-full             enable full talloc leak reporting on exit

Daemon options:
  -D, --daemon                       Become a daemon (default)
  -i, --interactive                  Run interactive (not a daemon) and log to stdout
  -F, --foreground                   Run daemon in foreground (for daemontools, etc.)
      --no-process-group             Don't create a new process group

Version options:
  -V, --version                      Print version
```

At the same time there are other(very few) options specific to each daemon. For example:

**smbd**
```
Usage: smbd [OPTION...]
  -b, --build-options                     Print build options
  -p, --port=STRING                       Listen on the specified ports
  -P, --profiling-level=PROFILE_LEVEL     Set profiling level
```
**winbindd**
```
Usage: winbindd [OPTION...]
  -n, --no-caching                   Disable caching
```
**nmbd**
```
Usage: nmbd [OPTION...]
  -H, --hosts=STRING                 Load a netbios hosts file
  -p, --port=INT                     Listen on the specified port
```

As we can see, `--debug-stdout` is common to these daemons and may be we can go for a single option `_DAEMON_CLI_STDOUT_OPT` or `_CLI_STDOUT_OPT`

phlogistonjohn (approved):

PR #19, opened 2021-10-06
enforce standard formatting

phlogistonjohn:
For a while I have been using [black](https://pypi.org/project/black/) for formatting the python code in sambacc. However, I never made it a formal part of the sambacc workflow. Now that sambacc is getting additional contributions, it's best to make the standard formatting black provides official.

These changes update a few files and adds a call to black to check (not reformat) the sambacc codebase when tox is executed.


anoopcs9 (approved):
lgtm.

PR #23, opened 2021-11-03
only send dns registrations when run on "leader" node

phlogistonjohn:
In some cases certain operations should only be handled by one "node" in a cluster. In the case of a samba container we want to prevent all the nodes in a cluster from submitting DNS registration when the state file changes, and only have the current leader handle the submission.

These changes add general infrastructure for determining a leader and then uses it to limit registration when ctdb is enabled.


spuiuk (approved):
Looks good. Ack!

PR #25, opened 2021-11-04
Replace some print calls with logging

phlogistonjohn:
In many cases, logging is preferable to simple print calls.
These patches replace some obvious cases where logging is better than prints.


phlogistonjohn:
Needs to be rebased on #23 to get the CI fix.

gd (approved):
LGTM, thanks!

Issue #21, opened 2021-10-14
commit 89477bd prevents checkout of branch


spuiuk:
Commit 
89477bd Git checkout to a branch to avoid detached HEAD
adds the following change

```
-        git checkout "${node}"
+        git checkout -b "${node}" "${node}"
```

to  tests/container/build.sh.

This is however problematic when we would like to build off a branch. 
eg: When building the container off the latest commit, we would like to the master branch. This results in the following command
git checkout -b "master" "master"
which fails with 
fatal: A branch named 'master' already exists.

The problem was seen when working on the samba-container PR
https://github.com/samba-in-kubernetes/samba-container/pull/52

PR #26, opened 2021-11-12
Revert `89477bd` to allow checking out any git refspec

anoopcs9:
Fixes: #21

### DIFF
--- a/tests/container/build.sh
+++ b/tests/container/build.sh
@@ -24,7 +24,7 @@ update() {
     if [[ -d "$d/.hg" ]]; then
         hg update --check "${node}"
     else
-        git checkout -b "${node}" "${node}"
+        git checkout "${node}"
     fi
 }
 


phlogistonjohn (commented):
I'm curious as to why you're avoiding the "disconnected head" checkout, and just disable the warning?

This seems a bit more complicated that what we had before you added the `-b` option.Is there some other downside to running `git checkout` the way we were before beyond the warning it prints?

anoopcs9:
> I'm curious as to why you're avoiding the "disconnected head" checkout, and just disable the warning?

Hm..somehow I didn't like it. But that's fine. We can live with the warning(or disable using some git config as you mentioned).

> This seems a bit more complicated that what we had before you added the `-b` option.Is there some other downside to running `git checkout` the way we were before beyond the warning it prints?

But we need to be able to checkout branch, tag and commit hash, right?

phlogistonjohn:
> > I'm curious as to why you're avoiding the "disconnected head" checkout, and just disable the warning?
> 
> Hm..somehow I didn't like it. But that's fine. We can live with the warning(or disable using some git config as you mentioned).
> 

I don't have any issues disabling the warning... and the warning text itself tells you how to disable it.  FWIW I check out commits into disconnected head state fairly often without issue. However, I'm always happy to learn more if there are other issues people run into.

> > This seems a bit more complicated that what we had before you added the `-b` option.Is there some other downside to running `git checkout` the way we were before beyond the warning it prints?
> 
> But we need to be able to checkout branch, tag and commit hash, right?

That's correct. AFAICT, `git checkout {branch/tag/hash}` worked for those cases.  You can't (simply) start making commits on top of the disconnected head in the case of the hash, but this checkout only exists in the container and can be (generally) treated as read-only.

anoopcs9:
> > > This seems a bit more complicated that what we had before you added the `-b` option.Is there some other downside to running `git checkout` the way we were before beyond the warning it prints?
> > 
> > But we need to be able to checkout branch, tag and commit hash, right?
> 
> That's correct. AFAICT, `git checkout {branch/tag/hash}` worked for those cases. You can't (simply) start making commits on top of the disconnected head in the case of the hash, but this checkout only exists in the container and can be (generally) treated as read-only.

Correct.
In that case, revert of 89477bd6e6d810845d6faf21120d439b49a373af would be sufficient here.

phlogistonjohn (commented):
I'll approve & merge this as-is if you'd like, but I'll leave this open if you want to disable the warning. If you think its not worth it, please just let me know and we can wrap this up.

anoopcs9:
> I'll approve & merge this as-is if you'd like, but I'll leave this open if you want to disable the warning. If you think its not worth it, please just let me know and we can wrap this up.

Just go ahead with the current revert.

phlogistonjohn (approved):
LGTM. Thanks!

PR #28, opened 2022-01-09
Initial AD DC Support

phlogistonjohn:
While it's a bit rough and simplistic, these changes add AD DC support to sambacc. Currently, there's one new script `samba-dc-container` that supports one subcommand `run`. It also has setup actions for `provision` and `join` and can also wait for the domain to appear with `wait-domain` which will poll until the ldap srv record exists for the domain.

With this sambacc supports (by default) the same behavior as the hard-coded scripts in samba-container's ad-server container image. In addition, you can customize your domain and initial users through the sambacc config json. You can then join additional containers to the domain.

I probably could have added more unit tests but the layer between us and samba-tool is thin currently. In the future, some of the stuff like user and groups creation may be better done via function calls into samba's python libs, but right now we just shell out to the cli.

### DIFF
--- a/examples/addc.json
+++ b/examples/addc.json
@@ -1,0 +1,127 @@
+{
+  "samba-container-config": "v0",
+  "configs": {
+    "demo": {
+      "instance_features": ["addc"],
+      "domain_settings": "sink",
+      "instance_name": "dc1"
+    }
+  },
+  "domain_settings": {
+    "sink": {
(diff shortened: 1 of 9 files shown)


phlogistonjohn:
@anoopcs9 @spuiuk  PTAL when you have a moment. While there's no huge rush for this I'd also like to make sure it doesn't linger for too long.These changes should not have any impact on the non-DC code.

PR #29, opened 2022-01-16
do not try to enable debugging for ldbtool

phlogistonjohn:
Fixes: #22

The main point for the PR is to fix the issue where ltdbtool fails when run with samba's standard approach for setting debug level on the cli. This tool is very simple and low level and thus is not built to accept that option. 
To make this change I factored a new base class out of SambaCommand and made our ltdbtool wrapper based on that.

While I was in the neighborhood I added tests for the new class as well some of the previously uncovered functions in samba_cmds.

### DIFF
--- a/sambacc/samba_cmds.py
+++ b/sambacc/samba_cmds.py
@@ -171,4 +171,4 @@ def encode(value: typing.Union[str, bytes, None]) -> bytes:
 
 def execute(cmd: SambaCommand) -> None:
     """Exec into the command specified (without forking)."""
-    os.execvp(cmd.name, cmd.argv())
+    os.execvp(cmd.name, cmd.argv())  # pragma: no cover
--- a/sambacc/samba_cmds.py
+++ b/sambacc/samba_cmds.py
@@ -35,8 +35,9 @@ _DAEMON_CLI_STDOUT_OPT: str = "daemon_cli_debug_output"
 def get_samba_specifics() -> typing.Set[str]:
     value = os.environ.get("SAMBA_SPECIFICS", "")
     out = set()
-    for v in value.split(","):
-        out.add(v)
+    if value:
+        for v in value.split(","):
+            out.add(v)
     return out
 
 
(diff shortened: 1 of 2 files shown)


sambacc/samba_cmds.py
```diff
@@ -35,8 +35,9 @@
 def get_samba_specifics() -> typing.Set[str]:
     value = os.environ.get("SAMBA_SPECIFICS", "")
     out = set()
-    for v in value.split(","):
-        out.add(v)
+    if value:
```
anoopcs9, line 38: Good catch.

sambacc/samba_cmds.py
```diff
@@ -60,44 +61,71 @@ def set_global_debug(level: str) -> None:
...
+
+    def _debug_args(self, dlvl: str = "--debuglevel={}") -> typing.List[str]:
+        if self.debug:
+            return [dlvl.format(self.debug)]
+        if _GLOBAL_DEBUG:
+            return [dlvl.format(_GLOBAL_DEBUG)]
+        return []
+
+    def raw_args(self):
+        return [self.name] + self._debug_args() + self.args
```
anoopcs9, line 127: Ok. So basically you add __debug_args()_ to _argv()_ by overriding _raw_args()_.

sambacc/samba_cmds.py
```diff
@@ -60,44 +61,71 @@ def set_global_debug(level: str) -> None:
...
+        )
+
+    def _debug_args(self, dlvl: str = "--debuglevel={}") -> typing.List[str]:
+        if self.debug:
+            return [dlvl.format(self.debug)]
+        if _GLOBAL_DEBUG:
+            return [dlvl.format(_GLOBAL_DEBUG)]
+        return []
+
+    def raw_args(self):
```
anoopcs9, line 126: Can't we  use return typing hint `typing.List[str]` for _raw_args()_?

anoopcs9 (commented):
I think I got the class split you made here to accept commands(with arguments) in general. Just a small nit(see inline comment).

Otherwise looks OK to me.

sambacc/samba_cmds.py
phlogistonjohn: Yes. I guess I just forgot it.

phlogistonjohn: fixed.

anoopcs9 (approved):
Thanks.

PR #30, opened 2022-02-09
commands: Add ctdb-check-ok

spuiuk:
This return the status of the ctdb daemon and is expected to be used as
the readiness probe by kubernetes.

Signed-off-by: Sachin Prabhu <sprabhu@redhat.com>


spuiuk:
Test:

Ensure ctdb status is ok
[root@smbshare3-0 /]# ctdb nodestatus
pnn:0 10.244.1.15      OK (THIS NODE)

Run the new command and check return condition.
[root@smbshare3-0 /]# samba-container ctdb-check-ok; echo $?
pnn:0 10.244.1.15      OK (THIS NODE)
0

Now disable and test the new command again with return condition.
[root@smbshare3-0 /]# ctdb disable
[root@smbshare3-0 /]# ctdb nodestatus
pnn:0 10.244.1.15      DISABLED (THIS NODE)
[root@smbshare3-0 /]# samba-container ctdb-check-ok; echo $?
pnn:0 10.244.1.15      DISABLED (THIS NODE)
4

sambacc/commands/ctdb.py
```diff
@@ -257,3 +257,10 @@ def ctdb_must_have_node(ctx: Context) -> None:
             return
         _logger.info("node not yet ready")
         waiter.wait()
+
+
+@commands.command(name="ctdb-check-ok")
+def ctdb_check_ok(ctx: Context) -> None:
+    """Check if ctdb service status is normal."""
+    _ctdb_ok()
+    ctdb.check_ok()
```
phlogistonjohn, line 266: This is not a problem with your PR, but I realized this call and the previous `_ctdb_ok()` call right after another is a tad bit confusing. However, I'm going to blame myself for choosing `_ctdb_ok` as the name for the function that checks the "feature gate"

sambacc/ctdb.py
```diff
@@ -483,6 +483,11 @@ def _convert_tdb_file(tdb_path: str, dest_dir: str, pnn: int = 0) -> None:
     subprocess.check_call(list(cmd))
 
 
+def check_ok(cmd: samba_cmds.SambaCommand = samba_cmds.ctdb):
+    cmd_ctdb_check = cmd["nodestatus"]
+    samba_cmds.execute(cmd_ctdb_check)
```
phlogistonjohn, line 488: When we discussed this before I thought we needed to read the stdout of the ctdb command. If so, we probably want to use a subprocess,  rather than execute, because execute does not fork.
If we don't need to read stdout, that's fine. I just want to understand better  what the expectation is.

tests/test_ctdb.py
```diff
@@ -563,3 +563,40 @@ def _fake_ctdb_script(pnn, recmaster):
...
+        'if [ "$1$TESTFAIL" == "nodestatus" ]',
+        "then exit 0;",
+        "else exit 1;",
+        "fi",
+    ]
+    fake_ctdb_script = datapath / "ctdb.sh"
+    with open(fake_ctdb_script, "w") as fh:
+        fh.write("\n".join(fake_ctdb))
+        fh.write("\n")
+    os.chmod(fake_ctdb_script, 0x0755)
```
phlogistonjohn, line 585: Generally, chmod is done using octal notation.
```suggestion    
    os.chmod(fake_ctdb_script, 0o0755)
```
If this is working it might be that were just lucky as 0x0755 != 0o755

phlogistonjohn (changes requested):
Thanks for looking into this, a few items inline:

tests/test_ctdb.py
spuiuk: baah brain fade!

sambacc/ctdb.py
spuiuk: No. We do not need to parse the output of the 'ctdb nodestatus' command. The return value of the ctdb nodestatus command  indicates the status of the ctdb service. Since the readiness probe depends solely on the return value, it made sense to just use exec() instead of fork() and exec()

sambacc/commands/ctdb.py
spuiuk: I was wondering if I should change the command to ctdb-status and change the corresponding function to ctdb_status(). or ctdb-nodestatus and ctdb_nodestatus()

phlogistonjohn: I think it's fine for now.

sambacc/ctdb.py
phlogistonjohn: OK.

tests/test_ctdb.py
spuiuk: Found the cause for my bug. Cut paste from testjoiner() in test_join.py. Added a new commit to the PR to fix that as well.

sambacc/commands/ctdb.py
```diff
@@ -257,3 +257,10 @@ def ctdb_must_have_node(ctx: Context) -> None:
             return
         _logger.info("node not yet ready")
         waiter.wait()
+
+
+@commands.command(name="ctdb-check-ok")
```
phlogistonjohn, line 262: We discussed in a private chat that this name could be improved. After a little more thinking I think we should mirror what is going on in `sambacc/commands/check.py` I suggest "ctdb-check" with a target value that is currently "node-status" (or "nodestatus" either is fine.)

phlogistonjohn (changes requested):

phlogistonjohn (approved):
LGTM, thanks for finding and fixing the existing chmod issue!

PR #36, opened 2022-03-25
Support updating samba config

phlogistonjohn:
Add support for updating the samba configuration. This can be done as a one time action with the command line `samba-container update-config`. The command also has a `--watch` option which causes the command to continuously loop waiting for changes to the configuration and then applying those to the samba/smbd config.

As part of the need to support adding shares to an existing container, this command also creates share dirs that do not already exist. Creating the share dirs is also now available as a `--setup=share_paths` setup step for the `run` subcommand. This is also available as a standalone subcommand `ensure-share-paths`.

A few general fixes and cleanups come along for the ride.

### DIFF
--- a/sambacc/commands/config.py
+++ b/sambacc/commands/config.py
@@ -16,12 +16,28 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>
 #
 
+import argparse
+import functools
+import logging
+import subprocess
 import sys
+import typing
 
+from sambacc import config
+from sambacc import samba_cmds
+from sambacc.simple_waiter import watch
 import sambacc.netcmd_loader as nc
 import sambacc.paths as paths
 
-from .cli import commands, setup_steps, Context
+from .cli import (
+    Context,
+    best_leader_locator,
+    best_waiter,
+    commands,
+    setup_steps,
+)
+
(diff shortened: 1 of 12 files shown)


spuiuk:
Two questions

It looks like we use the simple waiter by default. I was just wondering if we test inotify with a config map?  I guess it is a bind mount from the node filesystem and so should work but I am unsure if the underlying filesystem will support inotify.

Do we need smb-control in the command update-config? When using the registry shouldn't the latest config be autoloaded? Do we need to call smbcontrol?

phlogistonjohn:
> Two questions
> 
> It looks like we use the simple waiter by default. I was just wondering if we test inotify with a config map? I guess it is a bind mount from the node filesystem and so should work but I am unsure if the underlying filesystem will support inotify.

Good question, I do believe the config map is a bind mount plus a symlink to the target locations in the file system (IIRC).  I have not tested the inotify waiter in this scenario.  However, I think it's not currently enabled in our container images. I think fixing that is outside the scope of this PR.

> 
> Do we need smb-control in the command update-config? When using the registry shouldn't the latest config be autoloaded? Do we need to call smbcontrol?

@gd told me we should call it.

PR #38, opened 2022-05-03
More type annotations

phlogistonjohn:
To improve the type annotations usage in sambacc and hopefully catch more errors early the first step is to enable the option "disallow_incomplete_defs" in mypy. This disallows mixing typed and untyped parameters and return types on a single function call. It does not affect a function that has no annotations at all.

The first set of commits adds annotations to many parts of the code base. Then we enable the option in mypy to enforce it in the future (via pyproject.toml). Finally, we update a few straggling uses of typing.List and typing.Dict to the 'list' and 'dict' equivalents since our minimum version is python3.9.

### DIFF
--- a/sambacc/ctdb.py
+++ b/sambacc/ctdb.py
@@ -26,6 +26,7 @@ from sambacc import config
 from sambacc import jfile
 from sambacc import samba_cmds
 from sambacc.netcmd_loader import template_config
+from sambacc.typelets import ExcType, ExcValue, ExcTraceback
 
 _logger = logging.getLogger(__name__)
 
@@ -84,7 +85,9 @@ def ensure_ctdb_conf(
         write_ctdb_conf(fh, iconfig.ctdb_config())
 
 
-def write_ctdb_conf(fh: typing.IO, ctdb_params: typing.Dict, enc=str) -> None:
+def write_ctdb_conf(
+    fh: typing.IO, ctdb_params: typing.Dict, enc: typing.Callable = str
+) -> None:
     """Write a ctdb.conf style output."""
 
     def _write_param(fh: typing.IO, name: str, key: str) -> None:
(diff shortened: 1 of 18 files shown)


synarete (commented):
A general comment: PEP484, in [The meaning of annotations](https://peps.python.org/pep-0484/#the-meaning-of-annotations) requires `__init__` to have `-> None` type hinting. Not all cases follow this rule.

sambacc/ctdb.py
```diff
@@ -298,7 +312,9 @@ def _node_check(pnn: int, nodes_json: str, real_path: str) -> bool:
     return True
 
 
-def _node_update_check(json_data, nodes_json: str, real_path: str):
+def _node_update_check(
+    json_data: dict[str, typing.Any], nodes_json: str, real_path: str
```
synarete, line 316: `nodes_json` is not used in this function. With type hinting, functions signatures become longer; maybe it would be better to avoid used arguments.

phlogistonjohn:
> A general comment: PEP484, in [The meaning of annotations](https://peps.python.org/pep-0484/#the-meaning-of-annotations) requires `__init__` to have `-> None` type hinting. Not all cases follow this rule.

Right. This was (is?) tricky for me when I started using annotations.  My plan is to enable annotations required for `sambacc` but not `sambacc.commands` (or tests). This PR adds the first set of configuration params for mypy to the toml file. A future PR will add per-directory configs.

sambacc/ctdb.py
phlogistonjohn: Function is private so it should be straightforward.

phlogistonjohn:
Rebased.

synarete (approved):
LGTM

phlogistonjohn:
Thanks!!

PR #39, opened 2022-05-16
Update base container image to fedora 35

phlogistonjohn:
Fedora 34 is no longer supported with the recent release of F 36. 
With python 3.9 no longer being the "native" version of python I had to juggle a few things wrt to the tox test environments. We want to continue testing with python 3.9 in case we want to also support centos/rhel 9.
So we end up needing runs specifcically for py39 and some for whatever the system version of python is that can pull in native packaged libs like samba/inotify/xattr/etc.

This work is also a stepping stone to building an RPM so that the samba-container images can be "fully" RPMized and we can have a single source of truth for system python package deps (here in sambacc).

### DIFF
--- a/sambacc/_xattr.py
+++ b/sambacc/_xattr.py
@@ -1,0 +1,78 @@
+#
+# sambacc: a samba container configuration tool
+# Copyright (C) 2022  John Mulligan
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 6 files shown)


sambacc/_xattr.py
```diff
@@ -0,0 +1,78 @@
...
+could have been accomplished by writing a pyi file for xattr but since
+we need the runtime support we just add new functions.
+"""
+
+
+import pathlib
+import typing
+
+XAttrItem = typing.Union[
+    int,
```
synarete, line 35: I stared at this `int` for few minutes, thinking t myself "how could you pass int to xattr.get?" until i realized this int is an open file-descriptor :)
Code looks good. My only comment is to add a comment next to this int field (some thing like `# an open fd`)

phlogistonjohn: :-)  will do

phlogistonjohn: done.

synarete:
LGTM

spuiuk (approved):
Looks good. 

Acked-by: Sachin Prabhu <sprabhu@redhat.com>

PR #40, opened 2022-05-16
add rpm spec and build rpm in container

phlogistonjohn:
Depends on: #39 

Adds an rpm spec and rework the build.sh script in order to build rpms as part of our process. This RPM can be later consumed by the samba-container rather than the wheel. The wheel is great but going by container will be a lot more consistent with how the rest of the samba container images are constructed.

Leaving in draft for dependency/ci testing/whatnot reasons for now.


phlogistonjohn:
Found issue with rpm contents. Moving back to draft for now.

phlogistonjohn:
Issue resolved. See passing tests in samba-in-kubernetes/samba-container#71

tests/container/Containerfile
```diff
@@ -12,8 +12,19 @@ RUN yum install \
...
+    rpm-build \
+    'python3dist(flake8)' \
+    'python3dist(inotify-simple)' \
+    'python3dist(mypy)' \
+    'python3dist(pytest)' \
+    'python3dist(pytest-cov)' \
+    'python3dist(setuptools-scm)' \
+    'python3dist(setuptools-scm)' \
+    'python3dist(tox-current-env)' \
+    'python3dist(wheel)' \
```
anoopcs9, line 27: Are these requirements generated as per `%generate_buildrequires` and/or `%pyproject_buildrequires`?

anoopcs9, line 27: Are these requirements generated as per `%generate_buildrequires` and/or `%pyproject_buildrequires`?

contrib/python-sambacc.spec
```diff
@@ -0,0 +1,66 @@
...
+# set pversion for a customized python package version string
+%{?!pversion: %define pversion %{xversion}}
+# set rversion for a customized rpm version
+%{?!rversion: %define rversion %{xversion}}
+
+
+Name:           python-%{bname}
+Version:        %{rversion}
+Release:        1%{?dist}
+Summary:        Samba Container Configurator
```
anoopcs9, line 13: Current description seems more like `Summary`. We could have more explanatory para for `%description` down the line.

anoopcs9, line 13: Current description seems more like `Summary`. We could have more explanatory para for %description down the line.

anoopcs9 (commented):
Looks good in general..few minor/cosmetic comments.

Also how about renaming `contrib/` to `extras/` ?

contrib/python-sambacc.spec
```diff
@@ -0,0 +1,66 @@
+%global bname sambacc
+# set xversion to define the default version number
+%define xversion 0.1
+# set pversion for a customized python package version string
+%{?!pversion: %define pversion %{xversion}}
+# set rversion for a customized rpm version
+%{?!rversion: %define rversion %{xversion}}
+
+
+Name:           python-%{bname}
```
synarete, line 10: Why not `Name: python3-%{bname}` ?

contrib/python-sambacc.spec
```diff
@@ -0,0 +1,66 @@
...
+BuildArch:      noarch
+BuildRequires:  python3-devel
+
+%global _description %{expand:
+A library and CLI tools for managing Samba servers in containers.
+}
+
+%description %_description
+
+%package -n python3-%{bname}
```
synarete, line 30: I am not an export to rpm spec files, but I think that if you change `Name: python3-%{bname}` you will not need this `%package -n` entry in you spec file.

tests/container/build.sh
```diff
@@ -64,22 +144,12 @@ export WRITABLE_PASSWD=yes
 export NSS_WRAPPER_PASSWD=/etc/passwd
 export NSS_WRAPPER_GROUP=/etc/group
 
-# Run tox with sitepackages enabled to allow access to system installed samba
-# modules. The container env already provides us control over the env.
-info "running test suite with tox"
-tox
```
synarete, line 70: On my machine, I get the following warn when when building:
```
py3-mypy run-test: commands[0] | mypy sambacc tests
WARNING: test command found but not installed in testenv
  cmd: /usr/bin/mypy
  env: /var/tmp/build/sambacc/.tox/py3-mypy
Maybe you forgot to specify a dependency? See also the allowlist_externals envconfig setting.

DEPRECATION WARNING: this will be an error in tox 4 and above!
```

(thou I could not figure out why)

synarete (commented):
Looks good (though I have rather limited knowledge on rpm spec files).

contrib/python-sambacc.spec
anoopcs9: I hope its based on [Python Packaging Guidelines](https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/#_naming) from Fedora.

tests/container/build.sh
phlogistonjohn: Yeah, I think the solution is not to have mypy in the system packages, but we need it for rpm build... there are other things I/we could look into for rpm building - like 'mock' but as it was only a warning I left it to be solved later.

contrib/python-sambacc.spec
phlogistonjohn: Indeed. I think the idea is if that during the py2->py3 transition, and perhaps some future py3->py4 transition the source name is 'python-<blah>` but based on the build it gets named as python3-blah, python2-blah, etc.

contrib/python-sambacc.spec
phlogistonjohn: I took this from the template provided in the doc referenced in the commit msg. https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/

synarete: Interesting. There is so much new stuff there since the last time I read those docs..

contrib/python-sambacc.spec
```diff
@@ -0,0 +1,66 @@
...
+Summary: %{summary}
+# Distro requires that are technially optional for the lib
+Requires: python3-samba
+Requires: python3-pyxattr
+
+%description -n python3-%{bname}  %_description
+
+
+%prep
+%autosetup -p1 -n %{bname}-%{pversion}
```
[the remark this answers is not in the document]
phlogistonjohn: I took that directly from the doc https://docs.fedoraproject.org/en-US/packaging-guidelines/Python/
I don't actually know what it does.

tests/container/build.sh
```diff
@@ -33,27 +35,105 @@ update() {
...
+        if echo "$ver" | grep -q "+" ; then
+            rversion="$(echo "${ver}" | sed -e 's/\.dev/~/' -e 's/+/./')"
+        else
+            rversion="$ver"
+        fi
+        info "Using rpm-version=${rversion} pkg-version=${ver}"
+        rpmbuild --nocheck -ta \
+            -D"pversion ${ver}" -D"rversion ${rversion}" \
+            -D"_rpmdir ${distdir}/RPMS" \
+            -D"_srcrpmdir ${distdir}/SRPMS" \
```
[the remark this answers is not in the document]
phlogistonjohn: ok

tests/container/Containerfile
phlogistonjohn: Yes, they're generated automatically by reading the from tox.ini/setup.cfg/etc. I had hoped that using `-e py3` would cause it to leave off certain test-only deps like flake8 and mypy. However, I couldn't figure out how to do that.

contrib/python-sambacc.spec
```diff
@@ -0,0 +1,66 @@
...
+
+
+%files -n python3-%{bname} -f %{pyproject_files}
+%doc README.*
+%{_bindir}/samba-container
+%{_bindir}/samba-dc-container
+%{_datadir}/%{bname}/examples/
+
+
+%changelog
```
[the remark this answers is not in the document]
phlogistonjohn: You mean when we actually release? Or some placeholder in this PR?  If the latter, could you provide an example?

tests/container/build.sh
```diff
@@ -33,27 +35,105 @@ update() {
...
-if checked_out "${bdir}" ; then
-    info "repo already checked out"
-else
-    info "cloning sambacc repo"
-    clone "$url" "${bdir}"
-fi
+chk() {
+    for x in $tasks; do case "$1" in
+        "$x") "$1" ; return $? ;;
+    esac; done
```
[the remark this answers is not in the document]
phlogistonjohn: fixed

tests/container/build.sh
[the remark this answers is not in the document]
phlogistonjohn: done

phlogistonjohn:
I made some small tweaks to the spec, but don't really want to deviate much from what the packaging guidelines recommend without good justification.   I hope that makes sense.

contrib/python-sambacc.spec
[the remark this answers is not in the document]
anoopcs9: An example would be:
```
%changelog
Fri May 20 2022 John Mulligan <jmulligan@redhat.com - 0.1
- Initial spec file
```

contrib/python-sambacc.spec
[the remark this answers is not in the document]
anoopcs9: Package maintainers copy of RPM spec file may contain additional patches to fix issues after a particular release is made upstream in which case `-p1`(same meaning as for `patch` command line tool) will allow them to modify source with patches listed as
```
Patch1: <patch-file-name>
Patch2: <patch-file-name>
```
provided these patches exist. Even though it doesn't hurt there is not much sense in associating with `%autosetup` here.

extras/python-sambacc.spec
```diff
@@ -0,0 +1,67 @@
...
+# set pversion for a customized python package version string
+%{?!pversion: %define pversion %{xversion}}
+# set rversion for a customized rpm version
+%{?!rversion: %define rversion %{xversion}}
+
+
+Name:           python-%{bname}
+Version:        %{rversion}
+Release:        1%{?dist}
+Summary:        Samba Container Configurator
```
anoopcs9, line 13:
```
Summary: A library and CLI tools for managing Samba servers in containers.
```

extras/python-sambacc.spec
```diff
@@ -0,0 +1,67 @@
...
+
+BuildArch:      noarch
+BuildRequires:  python3-devel
+
+%global _description %{expand:
+A Python library and CLI tools intended to act as a bridge between a container
+environment and Samba servers and utilities.
+}
+
+%description %_description
```
anoopcs9, line 29:
```
%description
A Python library and CLI tools intended to act as a bridge between a container
environment and Samba servers and utilities. It aims to consolidate, coordinate and
automate all of the low level steps of setting up smbd, users, groups, and other
supporting components.
```

contrib/python-sambacc.spec
anoopcs9: See my suggestion on new changes.

contrib/python-sambacc.spec
```diff
@@ -0,0 +1,66 @@
...
+
+%description %_description
+
+%package -n python3-%{bname}
+Summary: %{summary}
+# Distro requires that are technially optional for the lib
+Requires: python3-samba
+Requires: python3-pyxattr
+
+%description -n python3-%{bname}  %_description
```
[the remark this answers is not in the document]
anoopcs9: See my suggestion on updated changes.

contrib/python-sambacc.spec
```diff
@@ -0,0 +1,66 @@
...
+License:        GPLv3+
+URL:            https://github.com/samba-in-kubernetes/sambacc
+# sambacc is not released yet so we're leaving off the url for now
+# once packaged and released we can update this field
+Source:         %{bname}-%{pversion}.tar.gz
+
+BuildArch:      noarch
+BuildRequires:  python3-devel
+
+%global _description %{expand:
```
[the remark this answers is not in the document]
anoopcs9: See my suggestion on updated changes.

contrib/python-sambacc.spec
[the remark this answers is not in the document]
phlogistonjohn: Sure, I understand the format. My question was if you expect a changelong entry for these "pre-release" builds or you expect an entry to be added as part of the release process?  Since we've discussed that the first real release is going to be v0.2  I can put in a "phony" entry for 0.1... but I think I'd prefer not to. Thoughts?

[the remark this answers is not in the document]
anoopcs9: I agree. Let's have an entry for 0.2 release.

extras/python-sambacc.spec
anoopcs9: Hm..let's move on with the current change.

anoopcs9 (approved):

phlogistonjohn:
Thank you very much!

Issue #43, opened 2022-06-01
Configuring globals does not appear to work


tunaranch:
### Discussed in https://github.com/samba-in-kubernetes/sambacc/discussions/42

<div type='discussions-op-text'>

<sup>Originally posted by **tunaranch** May 30, 2022</sup>
Hi everyone.

I'm trying to stand up a samba AD DC in minikube for testing, and I can't seem to modify smb.conf to disable the strong auth requirement.

Here's my Kubernetes resources, startup log, and generated smb.conf file

https://gist.github.com/09aed2659f5eeeaa176edd5f4d87e801

Have I made a mistake somewhere?</div>

tunaranch:
The problem affects both the ad and regular images, as described in https://github.com/samba-in-kubernetes/sambacc/discussions/42#discussioncomment-2860271

phlogistonjohn:
Looking at the code I am pretty confident that the bug lies in sambacc's addc layer. Global options that are not directly part of the parameters needed for provisioning are silently discarded.  This is misleading as nothing in the docs or structure of the config file indicates that this might be the case.

In the  AD DC path we can't directly write to smb.conf or use the registry backend, but we can pass `--option` parameters to the  `samba-tool domain provision` command.

I don't think the issue affects (file) server images. On the file server images we use the registry backend for samba. This means that only a minimal set of params will exist in smb.conf. To see what parameters have been set up for samba try the commands `net conf list` or `testparm`.

fsdrw08:
any update to this issue?

phlogistonjohn:
Hi @fsdrw08, I'm sorry to say I had started working on this and ran into to some issues between how I assumed samba-tool was going to work versus how it worked in practice, so the effort was more than I anticipated. It's helpful to know someone else is interested in this - perhaps I'll take another look at it soon.

PR #45, opened 2022-06-03
Pass global options to provision and join calls

phlogistonjohn:
Support taking global options defined in the configuration and passing them onto the samba-tools commands that will generate smb.conf.

Fixes: #43

### DIFF
--- a/sambacc/addc.py
+++ b/sambacc/addc.py
@@ -20,6 +20,7 @@ import logging
 import subprocess
 import typing
 
+from sambacc import config
 from sambacc import samba_cmds
 
 _logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


phlogistonjohn:
Unit tests are fine, but testing it practically I don't see the option making it into smb.conf, even though I've confirmed were passing the --option=foo=bar CLI arguments.  I'm going to leave this as a draft until I do some more practical tests and understand what samba-tool expects the options to be.

phlogistonjohn:
Getting the exact combination of CLI and smb.conf to work is proving tricky. I haven't given up on this tho.

obnoxxx (changes requested):
please rebase on current master and resolve conflicts.

phlogistonjohn:
I can, unfortunately the code does not function as intended. I haven't had time to follow up on this as I think we'll need to better understand / discuss the behavior of samba-tool with regards to default options.  JFYI

phlogistonjohn:
Closed because this approach just won't work and it's been over 6 months outstanding. There's still an open issue to work on this which is *not* being closed.

PR #47, opened 2022-06-25
sambacc: complete type annotations

phlogistonjohn:
The sambacc dir is basically a library and we want to ensure high quality and good documentation so configure mypy to require
type annotations on functions defined in sambacc. Update code in sambacc to add type annotations to any functions lacking them.

The sambacc.commands on the other hand are mean to be only run as part of our CLI tools so we'll be a bit less strict there.

### DIFF
--- a/sambacc/container_dns.py
+++ b/sambacc/container_dns.py
@@ -16,6 +16,8 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>
 #
 
+from __future__ import annotations
+
 import json
 import subprocess
 import typing
@@ -27,18 +29,22 @@ INTERNAL: str = "internal"
 
 
 class HostState:
-    def __init__(self, ref="", items=[]):
-        self.ref = ref
-        self.items = items
+    T = typing.TypeVar("T", bound="HostState")
+
+    def __init__(
+        self, ref: str = "", items: typing.Optional[list[HostInfo]] = None
+    ) -> None:
+        self.ref: str = ref
+        self.items: list[HostInfo] = items or []
 
     @classmethod
-    def from_dict(cls, d):
+    def from_dict(cls: typing.Type[T], d: dict[str, typing.Any]) -> T:
         return cls(
             ref=d["ref"],
             items=[HostInfo.from_dict(i) for i in d.get("items", [])],
         )
 
-    def __eq__(self, other):
+    def __eq__(self, other: typing.Any) -> bool:
         return (
             self.ref == other.ref
             and len(self.items) == len(other.items)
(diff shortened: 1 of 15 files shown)


sambacc/container_dns.py
```diff
@@ -27,18 +29,22 @@
...
 class HostState:
-    def __init__(self, ref="", items=[]):
+    T = typing.TypeVar("T", bound="HostState")
+
+    def __init__(
+        self, ref: str = "", items: typing.Optional[list[HostInfo]] = None
+    ) -> None:
         self.ref = ref
-        self.items = items
+        self.items = items or []
```
synarete, line 38: Consider having (align with other member-initialization in this PR):
```
       self.ref: str = ref
       self.items: list[HostInfo] = items or []
```

sambacc/inotify_waiter.py
```diff
@@ -85,7 +85,7 @@ def _get_events(self):
             and ((event.mask & _inotify.flags.CLOSE_WRITE) != 0)
         ]
 
-    def _wait(self):
+    def _wait(self) -> typing.Iterator[None]:
```
synarete, line 88: Why no just have a single `wait` function? What is the advantage of using `next(typing.Iterator[None])` ?

synarete (commented):
Two minor comments. Overall, LGTM.

phlogistonjohn:
Wow. Somehow I managed to forget about this PR. I will need to revisit this soon!

sambacc/container_dns.py
phlogistonjohn: updated.

sambacc/inotify_waiter.py
phlogistonjohn: At the time I must have felt implementing the function in terms of an iterator was easier or clearer. This PR only adds the proper type annotation to the function, it doesn't change the runtime behavior. Let's leave that for another time.

phlogistonjohn:
I finally updated this PR. wow.
@synarete @spuiuk Please revisit. I'd appreciate reviews within a week's time, so I'm putting on a virtual deadline of 2022-11-15. thanks!

sambacc/samba_cmds.py
```diff
@@ -159,19 +159,19 @@ def __repr__(self) -> str:
 samba_dc = SambaCommand("/usr/sbin/samba")
 
 
-def smbd_foreground():
+def smbd_foreground() -> SambaCommand:
     return smbd[
         "--foreground", _daemon_stdout_opt("smbd"), "--no-process-group"
     ]
```
synarete, line 165: IMHO, now that we have explicit return type, those lines look a bit enigmatic. I would consider (in a separate PR) to have more explicit code (also in other places below):

```
def smbd_foreground() -> SambaCommand:
    return SambaCommand("/usr/sbin/smbd",
        [ "--foreground", _daemon_stdout_opt("smbd"), "--no-process-group" ])
```

synarete (approved):
LGTM

PR #50, opened 2022-09-16
Use newly replaced `leader` admin command for `ctdb`

anoopcs9:
With Samba [v4.16](https://www.samba.org/samba/history/samba-4.16.0.html), `recmaster` administrative sub-command for `ctdb` was renamed to `leader`. Therefore we make use of already existing mechanism to detect the presence of `leader` admin command and execute it for determining the leader node.


phlogistonjohn (approved):
LGTM.

PR #54, opened 2022-11-16
tox.ini: avoid one tox warning when running py3-mypy

phlogistonjohn:
Rework how the tox.ini file is organized in order to remove a `command found but not installed in testenv` when running tox. We currently need at least one test run to be executed with system site packages enabled, but we don't need it that way for mypy.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -11,13 +11,9 @@ passenv =
 deps =
     pytest
     pytest-cov
-    mypy
     dnspython
-    py3: inotify_simple
-    py3: pyxattr
 commands =
     py.test -v tests --cov=sambacc --cov-report=html {posargs}
-sitepackages = py3: True
 
 [testenv:{py3,py39}-mypy]
 deps =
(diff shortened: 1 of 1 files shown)


synarete (approved):
LGTM

tox.ini
```diff
@@ -26,6 +22,20 @@ deps =
 commands =
     mypy sambacc tests
 
+[testenv:py3]
+# In order to run tests that rely on "system level" packages (samba,
+# xattr, etc.), and not have a lot of test skips, we have to enable the
+# sitepackages option. However when it is eanbled and you already have a tool
```
obnoxxx, line 28:
```suggestion
# sitepackages option. However when it is enabled and you already have a tool
```

obnoxxx (changes requested):
just a small irrelevant typo ...

tox.ini
phlogistonjohn: typo fixed. thanks!

obnoxxx (approved):
Thanks for fixing the typo ! I don't know enough about tox etc, to provide a knowledgeable review, but it looks clean enough and seems to work ... LGTM

PR #57, opened 2022-12-27
sambacc: re-format with black version 23.1a1

synarete:
Latest black version (23.1a1) requires extra empty line after one-liner comment. This, in turn, causes samba-container CI to fail.

Signed-off-by: Shachar Sharon <ssharon@redhat.com>


phlogistonjohn (approved):
LGTM, thanks!

PR #59, opened 2023-01-16
workflows: add commitlint workflow

obnoxxx:
components in .commitlinrc.json aretaken from the existing git log.

Signed-off-by: Michael Adam <obnox@samba.org>

### DIFF
--- a/.commitlintrc.json
+++ b/.commitlintrc.json
@@ -1,0 +1,32 @@
+{
+  "extends": [
+    "@commitlint/config-conventional"
+  ],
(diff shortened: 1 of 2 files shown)


obnoxxx:
I have put this on draft status since it was suggested in https://github.com/samba-in-kubernetes/samba-operator/pull/272#issuecomment-1384389886 to first resolve  the corresponding samba-operator PR

obnoxxx:
this is not needed any more since we have enabled gitlint instead by merging #75

PR #60, opened 2023-01-20
Switch to CentOS Stream 9 as base image

anoopcs9:


obnoxxx:
@anoopcs9  - why "do-not-merge"?

phlogistonjohn:
Seems pretty reasonable. My main concern is that with the change in base os we're changing the versions of python that get used for running the tests. That said we can probably take this and iterate from their either getting py3.10 (or up) on centos stream 9 _or_ making the containerfile flexible to run on either a fedora or centos base.

Therefore I'm leaning on taking this and then tweaking it but let me meditate on a little bit before approving.

anoopcs9:
@Mergifyio rebase

phlogistonjohn:
FWIW, I was working on something else and wanted to try f37 instead of f36. My workaround for the build process was to disable the rpm install and use the python wheel - which doesn't care about the python abi version.

phlogistonjohn (approved):
LGTM. I think I will want to make this work on multiple base distros but that can be done later. Thanks!

obnoxxx:
@anoopcs9  apologies: I merged despite the do-not-merge label because @phlogistonjohn  has approved....  what now? 

why has mergify not kicked in?

anoopcs9:
> @anoopcs9 apologies: I merged despite the do-not-merge label because @phlogistonjohn has approved.... what now?

Actually I had something else in mind while setting the `do-not-merge` label i.e, installing **python3-sambacc**(built with CentOS Stream 9) on Fedora 36. _samba-container_ is layered with _sambacc:latest_ and _fedora:36_ and thus I was not sure of any dependency issues/errors while installing **python3-sambacc** on Fedora 36. And now with this merged we encounter some installtion issues, see https://github.com/samba-in-kubernetes/samba-container/actions/runs/3982840129/jobs/6827649451
```

phlogistonjohn:
> @anoopcs9 apologies: I merged despite the do-not-merge label because @phlogistonjohn has approved.... what now?
> 
> why has mergify not kicked in?

Mergify didn't kick in because not all the conditions were met. It would have ignored Anoop's do-not-merge label because of a typo. But it still would not have merged because the rule, summarized in english is:
Merge any PR that has 2 or more approvals and no requested changes. If the submitter is a maintainer then merge if the PR has one approval and is over two weeks old. Currently the only maintainers are me (@phlogistonjohn ) and @obnoxxx.

 I was going to follow up on the miantainers thing later this week.

PR #61, opened 2023-01-20
github: add mergify configuration YAML file

phlogistonjohn:
Start trying out mergify on sambacc.
This is highly borrowed from the go-ceph repo, which in turn highly borrowed from the ceph-csi project's mergify configuration. It may not be optimal but hopefully it is useful.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,0 +1,67 @@
+---
+defaults:
+  actions:
+    queue:
+      name: default
+      method: rebase
+      update_method: rebase
+
+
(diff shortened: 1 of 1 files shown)


obnoxxx:
@phlogistonjohn , I don't know anything about mergify configuration but since this stems from two working examples and seems to encode essentiallywhat we already discussed , I am inclined to approve and merge so that we can try it ...

obnoxxx (approved):
not an edducated approval but this stems from proven examples. let's put it to use ...

so lgtm

PR #63, opened 2023-01-25
container: self contained build script

phlogistonjohn:
Right now, sambacc and samba-containers is in a pretty awkward situation. The base oses for the images are different and we're building RPMs that have the wrong ABI for the samba-container images. 

These changes pull nearly all the logic out of the continerfile and into build.sh so that it is easier to build and test and varying base OSes. If this is merged we have the ability to create multi-stage builds in samba-containers project hat pretty much only need to consume sambacc's container image OR even simply drop that and do something like curl build.sh from sambacc's github repo. Regardless of how we use it in the samba-containers repo I think this is helpful step to get out of a bad sitation and gain flexibility when executing tests.

----

This now makes it possible to more easily test and reuse the sambacc
test (and build) container image for different base OSes. For example:

```
podman build -t sambacc:mini36 --build-arg=SAMBACC_BASE_IMAGE=registry.fedoraproject.org/fedora:36 --build-arg=SAMBACC_MINIMAL=yes -f tests/container/Containerfile
podman run --rm -it sambacc:mini36

podman build -t sambacc:c9s -f tests/container/Containerfile

podman build -t sambacc:mini9 --build-arg=SAMBACC_MINIMAL=yes -f tests/container/Containerfile

podman build -t sambacc:fc37 --build-arg=SAMBACC_BASE_IMAGE=registry.fedoraproject.org/fedora:37 -f tests/container/Containerfile
```

### DIFF
--- a/tests/container/Containerfile
+++ b/tests/container/Containerfile
@@ -1,30 +1,14 @@
-FROM registry.fedoraproject.org/fedora:36
+ARG SAMBACC_BASE_IMAGE='registry.fedoraproject.org/fedora:36'
+FROM $SAMBACC_BASE_IMAGE
 
-RUN yum install \
-    -y --setopt=install_weak_deps=False \
-    git \
-    mercurial \
-    python-pip \
-    python-pip-wheel \
-    python-setuptools \
-    python-setuptools-wheel \
-    python-tox \
-    python3-samba \
-    python3-wheel \
-    python3-pyxattr \
-    python3-devel \
-    python3.9 \
-    samba-common-tools \
-    rpm-build \
-    'python3dist(flake8)' \
-    'python3dist(inotify-simple)' \
-    'python3dist(mypy)' \
-    'python3dist(pytest)' \
-    'python3dist(pytest-cov)' \
-    'python3dist(setuptools-scm)' \
-    'python3dist(tox-current-env)' \
-    'python3dist(wheel)' \
-    && yum clean all \
-    && true
+
 COPY build.sh /usr/local/bin/build.sh
+
+# Set SAMBACC_MINIMAL to yes to build a container that only contains the
+# build.sh script on top of the base image. When called, build.sh will
+# automatically install the dependency packages. Installing packages on every
+# run can be slow, especially if you are hacking on the code or tests, so we
+# install those dependencies proactively by default.
+ARG SAMBACC_MINIMAL=no
+RUN if [ "$SAMBACC_MINIMAL" != "yes" ]; then /usr/local/bin/build.sh --install ; fi
 ENTRYPOINT ["/usr/local/bin/build.sh"]
--- a/tests/container/build.sh
+++ b/tests/container/build.sh
@@ -79,6 +79,59 @@ setup_update() {
     fi
 }
 
+task_sys_deps() {
+    info "installing system packages"
+    OS_VER=$(source /etc/os-release && echo "${ID}-${VERSION_ID}")
+    case "${OS_VER}" in
+        centos*)
+            info "detected centos (stream): ${OS_VER}"
+            use_centos=true
+        ;;
+        fedora*)
+            info "detected fedora: ${OS_VER}"
+            use_centos=
+        ;;
+        *)
+            info "unknown platform: ${OS_VER}"
+            return 1
+        ;;
+    esac
+
+    yum_args=("--setopt=install_weak_deps=False")
+    pkgs=(\
+        git \
+        mercurial \
+        python-pip \
+        python-pip-wheel \
+        python-setuptools \
+        python-setuptools-wheel \
+        python-tox \
+        python3-samba \
+        python3-wheel \
+        python3-pyxattr \
+        python3-devel \
+        python3.9 \
+        samba-common-tools \
+        rpm-build \
+        'python3dist(flake8)' \
+        'python3dist(inotify-simple)' \
+        'python3dist(mypy)' \
+        'python3dist(pytest)' \
+        'python3dist(pytest-cov)' \
+        'python3dist(setuptools-scm)' \
+        'python3dist(tox-current-env)' \
+        'python3dist(wheel)' \
+    )
+
+    if [ "$use_centos" ]; then
+        yum install -y epel-release
+        yum_args=(--enablerepo=crb)
+        pkgs+=(pyproject-rpm-macros)
+    fi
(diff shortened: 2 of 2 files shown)


phlogistonjohn:
I forgot to mention I don't think there's a quick fix in samba-containers by simply converting everything that uses sambacc to centos:stream9 because AFAIK there's no dc package so we simply can't convert the ad dc image to centos at the moment. I will follow up this PR with a PR on samba-containers that let's us remain on fedora for the ad dc image.

tests/container/Containerfile
```diff
@@ -1,33 +1,14 @@
...
-    'python3dist(pytest-cov)' \
-    'python3dist(setuptools-scm)' \
-    'python3dist(tox-current-env)' \
-    'python3dist(wheel)' \
-    && yum clean all \
-    && true
 COPY build.sh /usr/local/bin/build.sh
+
+# Set SAMBACC_MINIMAL to yes to build a container that only contains the
+# build.sh script on top of the base image. When called build.sh will
```
obnoxxx, line 8:
```suggestion
# build.sh script on top of the base image. When called, build.sh will
```

phlogistonjohn: Fixed, thanks!

obnoxxx (dismissed):
lgtm, thanks!

anoopcs9:
> I will follow up this PR with a PR on samba-containers that let's us remain on fedora for the ad dc image.

How do we deal with different Samba package versions?

anoopcs9:
> > I will follow up this PR with a PR on samba-containers that let's us remain on fedora for the ad dc image.
> 
> How do we deal with different Samba package versions in server and ad-dc image?

To be more clearer, I was thinking on how those server and ad-dc images would be consumed in deploying samba-operator. If we plan to maintain same version this would mean that we create specific tags for sambacc and samba-container like fc36, c9s etc. In that case why don't we just live with what we had earlier i.e, by reverting https://github.com/samba-in-kubernetes/sambacc/pull/60?

phlogistonjohn:
> > > I will follow up this PR with a PR on samba-containers that let's us remain on fedora for the ad dc image.
> > 
> > 
> > How do we deal with different Samba package versions in server and ad-dc image?
> 
> To be more clearer, I was thinking on how those server and ad-dc images would be consumed in deploying samba-operator. If we plan to maintain same version this would mean that we create specific tags for sambacc and samba-container like fc36, c9s etc. In that case why don't we just live with what we had earlier i.e, by reverting #60?

I see making some progress to using centos as a base image worthwhile, even if we made a bit of a mess on our way there. The fact is, while I'd love the ad dc image to be something useful for a wide variety of production-like use cases, the ad dc image is "merely" used by our test frameworks for verifying AD related features in samba-operator.

Since samba-operator never deploys ad dc images, all of our "production ready" bits can be on centos. We can and should work towards getting ad dc on the same base, but I don't think we have to do it all at once.  I would much prefer to use the same base for everything, but since that doesn't work at this time something's gotta give.

anoopcs9:
> > > > I will follow up this PR with a PR on samba-containers that let's us remain on fedora for the ad dc image.
> > > 
> > > 
> > > How do we deal with different Samba package versions in server and ad-dc image?
> > 
> > 
> > To be more clearer, I was thinking on how those server and ad-dc images would be consumed in deploying samba-operator. If we plan to maintain same version this would mean that we create specific tags for sambacc and samba-container like fc36, c9s etc. In that case why don't we just live with what we had earlier i.e, by reverting #60?
> 
> I see making some progress to using centos as a base image worthwhile, even if we made a bit of a mess on our way there. The fact is, while I'd love the ad dc image to be something useful for a wide variety of production-like use cases, the ad dc image is "merely" used by our test frameworks for verifying AD related features in samba-operator at this time.
> 
> Since samba-operator never deploys ad dc images, all of our "production ready" bits can be on centos. We can and should work towards getting ad dc on the same base, but I don't think we have to do it all at once. I would much prefer to use the same base for everything, but since that doesn't work at this time something's gotta give.

Even in that case I don't know if this is the right approach. I made an attempt to make things work back again by allowing _el9_ build of **python3-sambacc** to be installed on f36. With some extra configuration it may succeed finally. I know that we then end up with f36 based server image which I am OK to live with for some more time.

obnoxxx (approved):
lgtm

tests/container/build.sh
```diff
@@ -150,7 +203,23 @@ export WRITABLE_PASSWD=yes
...
+if [[ "$1" = "--install" ]]; then
+    task_sys_deps
+    exit $?
+fi
+
+# if critical packages (currently just git) are missing we assume that
+# we need to automatically enable the task_sys_deps step.
+# this step is not enabled by default due to the overhead that updating
+# the dnf repos creates on the overall build time.
+if ! command -v git &>/dev/null ; then
```
anoopcs9, line 217: Isn't this equivalent to checking `SAMBACC_MINIMAL`?

phlogistonjohn: SAMBACC_MINIMAL is there to create a container that is basically just a base layer (os) + a layer containing the build script itself  (no build dependencies). This block is typically triggered *within* a minimal container OR in the case that the build script is copied from one container to another to install all the missing dependencies.  The idea here is that a minimal container can just act as a fancy bundle for including the build.sh in other contexts.  Example:

(untested & simplified but you should get the idea)
```
FROM sambacc:minimal AS sambacc
FROM quay.io/centos/centos:stream9 AS builder
COPY --from=sambacc /usr/local/bin/build.sh /tmp/build.sh
# actually build sambacc on centos
RUN /tmp/build.sh {...other args...}
FROM quay.io/centos/centos:stream9
RUN dnf install deps
COPY --from=builder /var/tmp/build/*.rpm  /tmp/rpms/
RUN dnf install sambacc
```

anoopcs9: May be I am not grasping it completely but I had a different thought of changing `SAMBACC_MINIMAL` to an environment variable(`ENV`) while building sambacc container and to act based on its value instead of checking for _git_ here. This is under an assumption that _git_ won't be included in any of the base OS images.

anoopcs9 (approved):
Forgot to mark the approval.

phlogistonjohn:
Thanks for the approval. I feel like I'm doing a poor job of explaining how we currently build sambacc & how I envision these changes being used. Maybe if we do a voice meeting soon I can try doing a better job then.

PR #64, opened 2023-01-26
Correct smbd log option

dmulder:
The samba-containers `make test` is failing
because smbd can't start when passed an invalid
parameter. The parameter should be
`--debug-stdout`, not `--log-stdout`.

Signed-off-by: David Mulder <dmulder@samba.org>

### DIFF
--- a/sambacc/samba_cmds.py
+++ b/sambacc/samba_cmds.py
@@ -44,7 +44,7 @@ def get_samba_specifics() -> typing.Set[str]:
 
 def _daemon_stdout_opt(daemon: str) -> str:
     if daemon == "smbd":
-        opt = "--log-stdout"
+        opt = "--debug-stdout"
     else:
         opt = "--stdout"
     opt_lst = get_samba_specifics()


phlogistonjohn:
First off, thanks very much for the PR. I'm always thrilled to se a new name on our repo.

sambacc has a feature of trying to support a few different versions of samba. These variations are controlled by an environment variable called SAMBA_SPECIFICS. You can look a few lines below the line you changed and see how the code changes the option used in reaction to the SAMBA_SPECIFICS value. I wonder if the better fix is to set the appropriate SAMBA_SPECIFICS value.

Also, could you share a link to the failing test? We're currently dealing with a bit of instability with regards to our base images and I may have overlooked this particular failure. 

cc: @anoopcs9

dmulder:
Ah, that makes sense. I'll take a look at SAMBA_SPECIFICS. I ran the test locally, and it's actually against an openSUSE image (you can see I'm working on our own images here: https://github.com/openSUSE/samba-container). So, there are bound to be a few differences.

phlogistonjohn:
OK, very interesting thanks.  FWIW, I'd be open to discussions about supporting different base images at samba-containers. Regardless of that, I think using or perhaps enhancing the SAMBA_SPECIFICS support would be the way to go.

dmulder:
Supporting different base images would be a great idea. I'm not sure what that will look like though.
Right now I'm building sambacc in our build system, then installing the package on the container. I noticed that you guys are building sambacc directly within the container when creating the image. Is this strictly necessary? Perhaps it has something to do with getting the samba version to set SAMBA_SPECIFICS?

dmulder:
Now that I look closer, I realize that get_samba_specifics() is already handling the '--debug-stdout' option. So I'll close this request.

PR #65, opened 2023-01-30
Revert "container: Switch to CentOS Stream 9 as base image"

phlogistonjohn:
This reverts commit af63900e4a0f2cef025656a7242c2cdda73fc845.

This should fix the current failures in samba-containers, allowing us to have a more in-depth discussion of what the right way forward is.

Signed-off-by: John Mulligan <jmulligan@redhat.com>


obnoxxx (approved):
lgtm, as discussed

PR #73, opened 2023-04-06
Add support for config files in TOML and YAML formats

phlogistonjohn:
Depends on: #69 

JSON is great in that it is standardized, tons of languages support it, and it's pretty simple to parse and work with. It is pretty terrible to edit by hand though. This PR adds support for both TOML and YAML as formats for our sambacc configuration.  The intent here is to make writing a config by hand simpler and thus using samba-container directly without management by the operator, for example, easier.

TOML is probably the nicest to work with once you memorize the rules around brackets (tables vs. arrays of tables).  This module is included in python, however, the python version must be 3.11 or later.

YAML is a superset of JSON that is much easier to write and is extremely popular, if a bit over-complicated.

The file format to use is selected based on the file name's extension. To use TOML the file name must end with ".toml". To use YAML the file name must end with either ".yml" or ".yaml". 

These changes build upon the work to use jsonschema. Despite the name the jsonschema can be applied to both these formats as it's really describing the underlying strucutre of the data (not the formatting) and both new formats are largely json-compatible, with similar structural and basic data types.

TODO:
* [x] Actually test it in a built samba container
* [x] Document the new formats with examples and descriptions
* [x] Update test deps to run (not skip) the newly added unit tests

### DIFF
--- a/sambacc/config.py
+++ b/sambacc/config.py
@@ -80,7 +80,13 @@ if sys.version_info >= (3, 11):
 else:
 
     def _load_toml(source: typing.IO) -> JSONData:
-        raise ConfigFormatUnsupported(ConfigFormat.TOML)
+        try:
+            import tomli
+        except ImportError:
+            raise ConfigFormatUnsupported(ConfigFormat.TOML)
+        if typing.TYPE_CHECKING:
+            assert isinstance(source, typing.BinaryIO)
+        return tomli.load(source)
 
 
 def _load_yaml(source: typing.IO) -> JSONData:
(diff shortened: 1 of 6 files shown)


sambacc/config.py
```diff
@@ -56,6 +63,31 @@ class _FakeRefResolutionError(Exception):
...
+else:
+
+    def _load_toml(source: typing.IO) -> JSONData:
+        raise ConfigFormatUnsupported(ConfigFormat.TOML)
+
+
+def _detect_format(fname: str) -> ConfigFormat:
+    if fname.endswith(".toml"):
+        return ConfigFormat.TOML
+    return ConfigFormat.JSON
```
synarete: OK, but I would prefer using  [splitext](https://docs.python.org/3/library/os.path.html#os.path.splitext) and compare with any of valid enum values of `ConfigFormat`. Just an idea.

synarete (approved):
LGTM

PR #74, opened 2023-04-06
fix rpm build issues introduced by jsonschema feature

phlogistonjohn:
Upon merging #69 the samba-container builds broke. 

The RPM build failed because the tox test dependencies listed jsonschema and no jsonschema python packages were installed. On top of that, the python3-jsonschema package in fedora 36 is incompatible and breaks the tests. So we're mostly just undoing the initial damage to try and get samba-containers working again.

To prevent future damage this PR enables the CI to run the RPM build step which was previously skipped (and only executed by samba-container builds).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
     - name: Build test container
       run: docker build  -t sambacc:ci  tests/container/ -f tests/container/Containerfile
     - name: Run test container
--- a/tox.ini
+++ b/tox.ini
@@ -12,7 +12,6 @@ deps =
     pytest
     pytest-cov
     dnspython
-    jsonschema
 commands =
     py.test -v tests --cov=sambacc --cov-report=html {posargs}
 
@@ -20,7 +19,6 @@ commands =
 deps =
     mypy
     types-setuptools
-    types-jsonschema
     types-pyyaml
     {[testenv]deps}
 commands =
--- a/sambacc/config.py
+++ b/sambacc/config.py
@@ -54,7 +54,7 @@ class ValidationUnsupported(Exception):
 
 def _schema_validate(data: dict[str, typing.Any], version: str) -> None:
     try:
-        import jsonschema
+        import jsonschema  # type: ignore[import]
     except ImportError:
         raise ValidationUnsupported()
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -13,7 +13,7 @@ url = https://github.com/samba-in-kubernetes/sambacc
 license = GPL3
 
 [options]
-packages = sambacc, sambacc.commands
+packages = sambacc, sambacc.commands, sambacc.schema
 include_package_data = True
 
 [options.entry_points]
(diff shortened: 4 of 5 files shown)


phlogistonjohn:
The build fails in the CI and works locally. Even though we're (in theory) using the same containerized environment.

The first thing that appears significantly different is:
```
 running egg_info
writing sambacc.egg-info/PKG-INFO
writing dependency_links to sambacc.egg-info/dependency_links.txt
writing entry points to sambacc.egg-info/entry_points.txt
writing top-level names to sambacc.egg-info/top_level.txt
listing git files failed - pretending there aren't any                       <---- this isn't in my output
reading manifest file 'sambacc.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
adding license file 'COPYING'
writing manifest file 'sambacc.egg-info/SOURCES.txt'
```

despite tweaking the checkout config, I still don't see why this happens

phlogistonjohn:
Fixed the problem, but now I need to clean up the debris.

.github/workflows/ci.yml
```diff
@@ -14,6 +14,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v3
+      with:
+        fetch-depth: 0
+        ref: ${{ github.event.pull_request.head.sha }}
```
anoopcs9, line 19: What happens to `ref` when test is triggered by either _push_ or _schedule_ event?

After all do we really need to use `ref` option? Because doc for the option says
```
When checking out the repository that triggered a workflow,
this defaults to the reference or SHA for that event.
```

tox.ini
```diff
@@ -12,15 +12,13 @@ deps =
     pytest
     pytest-cov
     dnspython
-    jsonschema
```
anoopcs9, line 15: IIUC we will reintroduce these changes(and the next one) once the required dependencies are sorted out within distribution, right?

.github/workflows/ci.yml
phlogistonjohn: Thanks for the reminder. For flows where we really need the ref we'll either need to use a different var or restrict it to pull request only. For this job I think it's merely a convenience because it makes the hashes embedded in the names match what you'd see if you build it locally. I'd like to leave it here for now because of that convenience and I don't think it'll break anything. If it does, I'll just jump in with another patch to remove it.

anoopcs9: For _push_ and _schedule_ events, `github.event.pull_request.head.sha` should be unknown. That's my concern. And sorry, I still don't understand the real need of `ref` option.

phlogistonjohn: There is no need AFAICT. I will push a new version without it. It does have an *advantage* in that the version numbers and git hashes that get added to the build exactly match those that you'd get if you build locally, but for this case it's not strictly necessary. Furthermore, if github.event.pull_request.head.sha is unavailable the action falls back to it's default, creating a merge commit, it doesn't break anything. Regardless, I will remove it. But you may see it reappear in other places where it actually does serve a significant purpose.

obnoxxx (approved):
LGTM

.github/workflows/ci.yml
anoopcs9: > Furthermore, if github.event.pull_request.head.sha is unavailable the action falls back to it's default, creating a merge commit, it doesn't break anything. Regardless, I will remove it. But you may see it reappear in other places where it actually does serve a significant purpose.

Ok.

PR #84, opened 2023-07-25
gitlint: ignore body lines that are footnotes

phlogistonjohn:
Ignore lines that start like `[1]: ` or `[2] ` such that we can have long text like URLs in the commits without triggering other gitlint body rules.

Copied from the same changes already made in samba-container (PR samba-in-kubernetes/samba-container#150).


synarete (approved):
lgtm

anoopcs9 (approved):

PR #85, opened 2023-08-11
add support for custom global options in AD DC confgurations

phlogistonjohn:
This revisits and reuses parts of the work done in #45
Fixes: #43 

When replying to samba-in-kubernetes/samba-container#154 I realized that one of the workarounds I suggested could be implemented in sambacc without too many complications and without needing to change `samba-tool`. Also, all versions of samba used by samba-container images that provide AD DC images should be providing the smbconf python library now. This avoids us having to parse smb.conf-style-ini in python.

Briefly, when an AD DC configuration provides globals references the global options are passed directory to samba-tool commands and after samba-tool has generated an smb.conf, sambacc uses smbconf libraries to read the smb.conf, merge the settings from the sambacc globals and write out an updated smb.conf file.

An example YAML configuration file I manually tested with:
```yaml
samba-container-config: v0
configs:
  demo:
    instance_features: [addc]
    domain_settings: sink
    instance_name: dc1
    globals: ["adglobals"]
domain_settings:
  sink:
    realm: DOMAIN2.SINK.TEST
    short_domain: DOMAIN2
    admin_password: Passw0rd
domain_groups:
  sink:
    - name: peeps
domain_users:
  sink:
    - name: bwayne
      password: "1115Rose."
      given_name: Bruce
      surname: Wayne
      member_of: ["peeps"]
globals:
  adglobals:
    options:
      "dns forwarder": "192.168.64.1"
      "ldap server require strong auth": "no"
      "ignore:me": "yes"
```

Results in an smb.conf like:
```
[global]
        dns forwarder = 192.168.64.1
        netbios name = dc1
        realm = DOMAIN2.SINK.TEST
        server role = active directory domain controller
        workgroup = DOMAIN2
        idmap_ldb:use rfc2307 = yes
        ignore:me = yes
        ldap server require strong auth = no

[sysvol]
        path = /var/lib/samba/sysvol
        read only = No

[netlogon]
        path = /var/lib/samba/sysvol/domain2.sink.test/scripts
        read only = No
```

Using `dig` in the container I was able to test that `dns forwarder` option did have an actual effect on the running `samba` server.

---

The PR starts off with a few random cleanups of stuff that I hit while working on these changes. If you'd rather see those as a separate PR, just ask.

This doesn't update any docs. I will update the documentation in a separate PR, perhaps after asking the issue reporters to test out the changes when they land in samba-container server & ad-server latest images on quay.

### DIFF
--- a/sambacc/addc.py
+++ b/sambacc/addc.py
@@ -31,6 +31,7 @@ def provision(
     admin_password: str,
     dns_backend: typing.Optional[str] = None,
     domain: typing.Optional[str] = None,
+    options: typing.Optional[typing.Iterable[tuple[str, str]]] = None,
 ) -> None:
     # this function is a direct translation of a previous shell script
     # as samba-tool is based on python libs, this function could possibly
@@ -43,6 +44,7 @@ def provision(
             admin_password=admin_password,
             dns_backend=dns_backend,
             domain=domain,
+            options=options,
         )
     )
     return
@@ -95,6 +97,7 @@ def _provision_cmd(
     admin_password: str,
     dns_backend: typing.Optional[str] = None,
     domain: typing.Optional[str] = None,
+    options: typing.Optional[typing.Iterable[tuple[str, str]]] = None,
 ) -> list[str]:
     if not dns_backend:
         dns_backend = "SAMBA_INTERNAL"
@@ -110,8 +113,12 @@ def _provision_cmd(
         f"--realm={realm}",
         f"--domain={domain}",
         f"--adminpass={admin_password}",
-    ].argv()
-    return cmd
+    ]
+    for okey, oval in options or []:
+        if okey == "netbios name":
+            continue
+        cmd = cmd[f"--option={okey}={oval}"]
+    return cmd.argv()
 
 
 def _join_cmd(
--- a/sambacc/addc.py
+++ b/sambacc/addc.py
@@ -56,6 +56,7 @@ def join(
     admin_password: str,
     dns_backend: typing.Optional[str] = None,
     domain: typing.Optional[str] = None,
+    options: typing.Optional[typing.Iterable[tuple[str, str]]] = None,
 ) -> None:
     _logger.info(f"Joining AD domain: realm={realm}")
     subprocess.check_call(
@@ -64,6 +65,7 @@ def join(
             dcname,
             admin_password=admin_password,
             dns_backend=dns_backend,
+            options=options,
         )
     )
 
@@ -127,6 +129,7 @@ def _join_cmd(
     admin_password: str,
     dns_backend: typing.Optional[str] = None,
     domain: typing.Optional[str] = None,
+    options: typing.Optional[typing.Iterable[tuple[str, str]]] = None,
 ) -> list[str]:
     if not dns_backend:
         dns_backend = "SAMBA_INTERNAL"
@@ -141,8 +144,12 @@ def _join_cmd(
         f"--option=netbios name={dcname}",
         f"--dns-backend={dns_backend}",
         f"--password={admin_password}",
-    ].argv()
-    return cmd
+    ]
+    for okey, oval in options or []:
+        if okey == "netbios name":
+            continue
+        cmd = cmd[f"--option={okey}={oval}"]
+    return cmd.argv()
 
 
 def _user_create_cmd(
--- a/sambacc/addc.py
+++ b/sambacc/addc.py
@@ -93,6 +93,14 @@ def add_group_members(group_name: str, members: list[str]) -> None:
     subprocess.check_call(cmd)
 
 
+def _filter_opts(
+    options: typing.Optional[typing.Iterable[tuple[str, str]]]
+) -> list[tuple[str, str]]:
+    _skip_keys = ["netbios name"]
+    options = options or []
+    return [(k, v) for (k, v) in options if k not in _skip_keys]
+
+
 def _provision_cmd(
     realm: str,
     dcname: str,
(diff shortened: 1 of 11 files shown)


tests/test_addc.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def _fake_samba_tool(path):
+    fake_samba_tool = path / "fake_samba_tool.sh"
+    with open(fake_samba_tool, "w") as fh:
+        fh.write("#!/bin/sh\n")
+        fh.write(f"[ -e {path}/fail ] && exit 1\n")
+        fh.write(f'echo "$@" > {path}/args.out\n')
+        fh.write("exit 0")
+    os.chmod(fake_samba_tool, 0o700)
```
synarete, line 34: The code is OK, but as you are already using modern Python `pathlib.Path`, wouldn't it be better to use its provided methods? In this case, [write_text](https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_text) and [chmod](https://docs.python.org/3/library/pathlib.html#pathlib.Path.chmod).

This comment holds for the rest of the code in this file.

sambacc/config.py
```diff
@@ -237,7 +237,11 @@ def __init__(self, conf: GlobalConfig, iconfig: dict):
     def global_options(self) -> typing.Iterable[typing.Tuple[str, str]]:
         """Iterate over global options."""
         # Pull in all global sections that apply
-        gnames = self.iconfig["globals"]
+        try:
+            gnames = self.iconfig["globals"]
+        except KeyError:
+            # no globals section in the instance means no global options
+            return
```
synarete, line 244: I say: `gnames = self.iconfig.get("globals", [])`

tests/test_smbconf_samba.py
```diff
@@ -0,0 +1,189 @@
...
+    except ImportError:
+        pytest.skip("unable to load samba smbconf modules")
+
+
+def _smb_data(path, smb_conf_text):
+    data_path = path / "_samba"
+    data_path.mkdir()
+    smb_conf_path = path / "smb.conf"
+    with open(smb_conf_path, "w") as fh:
+        fh.write(smb_conf_text.format(path=data_path))
```
synarete, line 67: Consider using `smb_conf_path.write_text(...)`

tests/test_smbconf_samba.py
```diff
@@ -0,0 +1,189 @@
...
+    return smb_conf_path
+
+
+@pytest.fixture(scope="session")
+def smbconf_reg_once(tmp_path_factory):
+    _import_probe()
+    tmp_path = tmp_path_factory.mktemp("smb_reg")
+    smb_conf_path = _smb_data(tmp_path, smb_conf_reg_stub)
+
+    return sambacc.smbconf_samba.SMBConf.from_registry(str(smb_conf_path))
```
synarete, line 77: Is the `str` of `smb_conf_path` really needed?

sambacc/commands/addc.py
```diff
@@ -105,6 +109,27 @@ def _prep_join(ctx: Context) -> None:
         admin_password=domconfig.admin_password,
         options=ctx.instance_config.global_options(),
     )
+    _merge_config(_provisioned, ctx.instance_config.global_options())
+
+
+def _merge_config(
+    smb_conf_path: str,
```
synarete, line 116: Consider using `pathlib.Path` instead

synarete (commented):
Overall, look fine. I would prefer modern `pathlib.Path` over traditional `os` APIs.

tests/test_addc.py
phlogistonjohn: I've been writing python for a long time, much longer than pathlib's been around. While I generally like pathlib, old habits die hard. The library is well integrated with the os functions. Plus, in the case of a more complex multi-line file I don't really see write_text as an advantage.  For the test code already using pathlib, I will try to use it more. However, I don't think I want to change this PR so extensively to do so.

sambacc/config.py
phlogistonjohn: This was an intentional choice in this case. It nets me a comment and an early return rather than having to fall thru to an empty loop or bother with an if statement.

tests/test_smbconf_samba.py
phlogistonjohn: Yes. Another reason I don't use pathlib much in the library code (as opposed to tests) is that functions provided by Samba do not take inputs other than strings, they're implemented in simple(r) C code and don't support anything else.

tests/test_addc.py
synarete: Fair enough.

sambacc/config.py
synarete: I now see it. Fair enough.

tests/test_smbconf_samba.py
phlogistonjohn: done

synarete (approved):
lgtm

sambacc/addc.py
```diff
@@ -89,12 +93,21 @@ def add_group_members(group_name: str, members: list[str]) -> None:
     subprocess.check_call(cmd)
 
 
+def _filter_opts(
+    options: typing.Optional[typing.Iterable[tuple[str, str]]]
+) -> list[tuple[str, str]]:
+    _skip_keys = ["netbios name"]
```
anoopcs9, line 99: I hope this is because we already accept **netbios name** as `instance_name` in the config.

anoopcs9 (commented):
We already support custom options in non AD DC setup, right?

sambacc/addc.py
phlogistonjohn: That's correct.

anoopcs9 (approved):

PR #91, opened 2023-09-19
[RFC] Support configuration from remote resources

phlogistonjohn:
Add initial support for sourcing configuration from remote URLs (URIs), in addition to the existing support for local paths.

Currently supports HTTP and HTTPS urls and optionally "pseudo URLs" for resources stored in Ceph RADOS. The latter look like "rados://pool/ns/obj_key".

If the rados library for python is not present sambacc will skip enabling rados support.

Previously, the configuration paths to sambacc could be passed by `--config` cli options or  the `SAMBACC_CONFIG` environment variable. In the case of the latter we accepted a colon-separated list of paths in the style of `PATH`. However this doesn't work well for uris so it now optionally accpts a JSON-formatted list of strings in the following style:
```
$ export SAMBACC_CONFIG='["http://localhost:8080/my/config.json", "/my/config.json"]'
$ samba-container --identity demo <...>
```

Using JSON is detected by a square bracket as the first and last(ish) character of the env var (trailing whitespace is allowed)'

## TODO
* Needs more manual testing for RADOS, etc.
* Extend opener support to join files


phlogistonjohn:
I finally had some time to retest the rados support manually. Looked OK. Ready for review.

synarete (approved):
LGTM

(A note for the future: better split such large PR into multiple small-PRs in order to simplify review).

PR #98, opened 2023-12-19
sambacc: add support for selecting ceph auth user/entity/id when reading config

phlogistonjohn:
Add a command line option --ceph-id and environment variable SAMBACC_CEPH_ID that allow one to specify the user/id/entity name (ceph is really inconsistent about what to call these IMO) that should be used to read the config. Annoyingly, ceph libraries default to using the admin user with maximum capabilities and so it's critical to remember to get this working for non-admin users - as this PR tries to do. 

The value provided to --ceph-id / SAMBACC_CEPH_ID can be either just the name string (like: `bob` or `client.bob`) or a key value pair like `name=client.bob`. The former is implemented for convenience but ceph tooling can be a bit ambiguous about if it expects/provides the prefixed (`client.`) or unprefixed form of names. The key-value form allows one to more precisely state if the name is "fully qualified" (`name=client.bob`) or not (`rados_id=bob`). The plain name-only form has to guess and just assumes anything starting with `client.` is already fully-qualified.

While I was working on this I hit <s>two</s> three additional items:
* First, there was no way for me to turn on debug logging so I added a --debug cli option.
* Second, there seems to be a behavior change in the newer version(s) of `black` so I temporarily capped the maximum version so it doesn't fail the CI on files that we haven't changed. In a future PR I will revisist this and fix things up in a nicer way.
* IIIrd, schemacheck on f39 would not run due to an indirect dep failing to install. Added a workaround

### DIFF
--- a/sambacc/rados_opener.py
+++ b/sambacc/rados_opener.py
@@ -27,6 +27,7 @@ from . import url_opener
 from .typelets import ExcType, ExcValue, ExcTraceback
 
 _RADOSModule = typing.Any
+_RADOSObject = typing.Any
 
 _CHUNK_SIZE = 4 * 1024
 
@@ -35,15 +36,33 @@ class RADOSUnsupported(Exception):
     pass
 
 
+class _RADOSInterface:
+    api: _RADOSModule
+    client_name: str
+    full_name: bool
+
+    def Rados(self) -> _RADOSObject:
+        name = rados_id = ""
+        if self.full_name:
+            name = self.client_name
+        else:
+            rados_id = self.client_name
+        return self.api.Rados(
+            name=name,
+            rados_id=rados_id,
+            conffile=self.api.Rados.DEFAULT_CONF_FILES,
+        )
+
+
 class _RADOSHandler(urllib.request.BaseHandler):
-    _rados_api: typing.Optional[_RADOSModule] = None
+    _interface: typing.Optional[_RADOSInterface] = None
 
     def rados_open(self, req: urllib.request.Request) -> typing.IO:
-        if self._rados_api is None:
+        if self._interface is None:
             raise RADOSUnsupported()
         if req.selector.startswith("mon-config-key:"):
             return _get_mon_config_key(
-                self._rados_api, req.selector.split(":", 1)[1]
+                self._interface, req.selector.split(":", 1)[1]
             )
         sel = req.selector.lstrip("/")
         if req.host:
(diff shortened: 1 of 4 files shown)


phlogistonjohn:
*sigh* I have no idea why pip on f39 is failing. It doesn't seem to be related to my changes but seems to be consistently failing ATM.

phlogistonjohn:
Added a workaround for the dependency/pip install issue no fedora 39.

sambacc/rados_opener.py
```diff
@@ -20,58 +20,80 @@
...
+
 
 class RADOSUnsupported(Exception):
     pass
 
 
+class _RADOSInterface:
+    api: _RADOSModule
+    client_name: str
+    full_name: bool
```
synarete, line 45: Why not have `__init__` function with explicit types? IMHO it would make the code more readable.

sambacc/rados_opener.py
```diff
@@ -209,7 +236,17 @@ def enable_rados_url_opener(cls: typing.Type[url_opener.URLOpener]) -> None:
...
 
-    _RADOSHandler._rados_api = rados
+    _logger.debug(
+        "Enabling ceph rados support with"
+        f" client_name={client_name!r}, full_name={full_name}"
+    )
+    rados_interface = _RADOSInterface()
+    rados_interface.api = rados
+    rados_interface.client_name = client_name
+    rados_interface.full_name = full_name
```
synarete, line 249: See above comment on `__init__` function. IMHO it would make code more readable here as well.

synarete (commented):
Looks fine. See minor comments.

sambacc/rados_opener.py
phlogistonjohn: Partly lazyness but also partly that I wanted to highlight the 'internal-ness' of this class. It should only be created inside other code that's already familiar with the structure. Do you still want it?

synarete: No need. It is just a matter of style.

synarete (approved):
lgtm

phlogistonjohn:
@anoopcs9 ping.

tox.ini
```diff
@@ -50,17 +50,25 @@ allowlist_externals =
...
 commands =
     python -m sambacc.schema.tool
+# The following env var is a workaround for running this testenv on fedora 39
+# with python 3.12.1. As of approx. 2023-12-11 the ci started # to fail because
+# of the multidict library (not even one of our direct dependencies) failing
+# to install from source because we (intentionally) have no C compiler in the
+# image. Try to remove this in a few weeks to see if wheel availability or
+# whatever has been resolved.
+setenv =
+    MULTIDICT_NO_EXTENSIONS=1
```
anoopcs9, line 71: From recent CI runs I could see failures with Fedora 38 too. Is that also related to 'multidict' library?

anoopcs9 (commented):
Looks good to me.

tox.ini
phlogistonjohn: Black formatting. This pr has a temorary fix for that until I take a look at what changed in the recent releases of black.

anoopcs9: Ah..Ok.

anoopcs9 (approved):

PR #102, opened 2024-01-03
explicitly use black version from 2023

phlogistonjohn:
The CI was failing recently due to is (somehow) using the black 24 alpha. This was hackily fixed in PR #98 and this PR more cleanly pins the project to use black's 2023 style for now.

I did find some obvious cleanups using the --preview (2024) style and took them in the other change in this pr. However, I'm not planning to commit to the 2024 style until understand better why it is "messing up" some of the protocol function definitions.


synarete (approved):
lgtm

PR #106, opened 2024-02-03
Fixed environment variables SAMBA_DEBUG_LEVEL and SAMBA_SPECIFICS behavior

### LINKED ISSUE #105: SAMBA_DEBUG_LEVEL and SAMBA_SPECIFICS environment variables don't work as expected
Failed to run a container with debug env var
```shell
~ podman run --rm -ti --privileged -e SAMBA_DEBUG_LEVEL=3 quay.io/samba.org/samba-ad-server
...
samba-tool: no such subcommand: --debuglevel=3
```

Ignore `SAMBA_SPECIFICS=daemon_cli_debug_output` env var for /usr/sbin/samba, but samba has this option
```py
def samba_dc_foreground() -> SambaCommand:
    return samba_dc["--foreground"]
```

myback:
fix #105

### DIFF
--- a/sambacc/samba_cmds.py
+++ b/sambacc/samba_cmds.py
@@ -140,7 +140,7 @@ class SambaCommand(CommandArgs):
         return []
 
     def raw_args(self) -> list[str]:
-        return [self._name] + self._debug_args() + self.args
+        return [self._name] + self.args + self._debug_args()
 
     def __repr__(self) -> str:
         return "SambaCommand({!r}, {!r}, {!r})".format(
--- a/sambacc/samba_cmds.py
+++ b/sambacc/samba_cmds.py
@@ -172,7 +172,7 @@ def winbindd_foreground() -> SambaCommand:
 
 
 def samba_dc_foreground() -> SambaCommand:
-    return samba_dc["--foreground"]
+    return samba_dc["--foreground", _daemon_stdout_opt("samba")]
 
 
 ctdbd = SambaCommand("/usr/sbin/ctdbd")


phlogistonjohn (approved):
Thanks. Looks OK to me.

sambacc/samba_cmds.py
```diff
@@ -172,7 +172,10 @@ def winbindd_foreground() -> SambaCommand:
 
 
 def samba_dc_foreground() -> SambaCommand:
-    return samba_dc["--foreground"]
+    samba_foreground = samba_dc["--foreground"]
+    if _DAEMON_CLI_STDOUT_OPT in get_samba_specifics():
+        samba_foreground = samba_foreground["--debug-stdout"]
+    return samba_foreground
```
anoopcs9, line 178: Do you foresee any problem with the following suggestion(similar to _smbd_ and _winbindd_)?

```suggestion
    return samba_dc["--foreground", _daemon_stdout_opt("samba")]
```

anoopcs9 (changes requested):

sambacc/samba_cmds.py
myback: without the SAMBA_SPECIFICS=daemon_cli_debug_output you will get error `Invalid option --stdout: unknown option`

myback: @anoopcs9 it works with the `SAMBA_SPECIFICS` variable set and have `daemon_cli_debug_output` value.
_daemon_stdout_opt("samba") function return `--stdout` if the `SAMBA_SPECIFICS` variable is not set or has a value other than `daemon_cli_debug_output`. The Samba binary has no `--stdout` arguments and fails.

```py
def _daemon_stdout_opt(daemon: str) -> str:
    if daemon == "smbd":
        opt = "--log-stdout"
    else:
        opt = "--stdout"
    opt_lst = get_samba_specifics()
    if _DAEMON_CLI_STDOUT_OPT in opt_lst:
        opt = "--debug-stdout"
    return opt
```

anoopcs9: Just now realized that corresponding [Containerfile](https://github.com/samba-in-kubernetes/samba-container/blob/master/images/ad-server/Containerfile.centos) for ad-server fails to update `SAMBA_SPECIFICS` during build process.

So the right solution would be to use _daemon_stdout_opt()_ here and subsequently fix the build.

myback: something like this?

anoopcs9: I don't think we have to modify _daemon_stdout_opt()_. Just the change from my earlier [suggestion](https://github.com/samba-in-kubernetes/sambacc/pull/106#discussion_r1477726791) is enough. In addition `samba-ad-server` image build process can be updated to set `SAMBA_SPECIFICS` from [samba-container](https://github.com/samba-in-kubernetes/samba-container) repository. Ultimately that's the container image in use here.

anoopcs9: > In addition `samba-ad-server` image build process can be updated to set `SAMBA_SPECIFICS` from [samba-container](https://github.com/samba-in-kubernetes/samba-container) repository.

https://github.com/samba-in-kubernetes/samba-container/commit/6100b0f49f9dcedaa5da58c85cb49b2c29f9ca20

phlogistonjohn (approved):
Looks OK to me. Thanks!

anoopcs9 (approved):
lgtm, thanks.

PR #109, opened 2024-03-02
addc: dynamic interface filtering for provisioning

phlogistonjohn:
Lately I keep hitting hosts that have "bogus" interfaces, stuff like "docker0" that are not suitable for including in the DNS records of the DC. While one could configuring these statically if you know all "good" interfaces ahead of time I keep bumping into the fact that I don't - nor do I want to carefully examine every host. What I do know is stuff like `docker0` or `virbirX` are often not suitable.

Thus I'm adding some new options for the domain_settings section. Under the domain one can have "interfaces" and interfaces can contain any of "include_pattern" and "exclude_pattern". Both take regexes and filter as they are named.  This way I can just start doing stuff like setting:

```
  "domain_settings": {
    "sink": {
      "realm": "DOMAIN1.SINK.TEST",
      "short_domain": "DOMAIN1",
      "admin_password": "Passw0rd",
      "interfaces": {
            "exclude_pattern": "^(virbr|docker).*$"
        }
    } 
  }, 
```

Interfaces is a subsection because I thought it looked neater and I figured there's a chance we'll want to extend the ways we filter interfaces in the future so it made sense to encapsulate it a bit.

### DIFF
--- a/sambacc/addc.py
+++ b/sambacc/addc.py
@@ -17,9 +17,11 @@
 #
 
 import logging
+import re
 import subprocess
 import typing
 
+from sambacc import config
 from sambacc import samba_cmds
 
 _logger = logging.getLogger(__name__)
@@ -215,3 +217,24 @@ def _group_add_members_cmd(group_name: str, members: list[str]) -> list[str]:
         ",".join(members),
     ].argv()
     return cmd
+
+
+def _ifnames() -> list[str]:
+    import socket
+
+    return [iface for _, iface in socket.if_nameindex()]
+
+
+def filtered_interfaces(
+    ic: config.DCInterfaceConfig, ifnames: typing.Optional[list[str]] = None
+) -> list[str]:
+    _include = re.compile(ic.include_pattern or "^.*$")
+    _exclude = re.compile(ic.exclude_pattern or "^$")
+    if ifnames is None:
+        ifnames = _ifnames()
+    return [
+        name
+        for name in ifnames
+        if (name == "lo")
+        or (_include.match(name) and not _exclude.match(name))
+    ]
--- a/sambacc/schema/conf-v0.schema.json
+++ b/sambacc/schema/conf-v0.schema.json
@@ -256,6 +256,19 @@
           },
           "admin_password": {
             "type": "string"
+          },
+          "interfaces": {
+            "type": "object",
+            "properties": {
+              "include_pattern": {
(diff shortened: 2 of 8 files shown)


phlogistonjohn:
I built a container with this sambacc branch and did a manual test on one of the troubelsome nodes. It worked as intended, so I'm moving it out of draft.

phlogistonjohn:
FWIW, I tried to avoid writing this code by just temporarily working around issues running a DC in the ceph sepia nodes, but I found I couldn't do so reliably, so I have resorted to code. :-)

sambacc/addc.py
```diff
@@ -215,3 +217,24 @@ def _group_add_members_cmd(group_name: str, members: list[str]) -> list[str]:
         ",".join(members),
     ].argv()
     return cmd
+
+
+def _ifnames() -> list[str]:
+    import socket
```
synarete, line 223: Why local import instead top-level one? 
(Not an issue, but wondering why have it like this).

synarete (approved):
lgtm

sambacc/addc.py
phlogistonjohn: I figured it would save an import at start-up time. Only if the interface filtering gets used would would the module get imported. It's probably a minuscule difference though. :-)

sambacc/schema/conf-v0.schema.json
```diff
@@ -256,6 +256,19 @@
           },
           "admin_password": {
             "type": "string"
+          },
+          "interfaces": {
+            "type": "object",
+            "properties": {
+              "include_pattern": {
+                "type": "string",
+                "description": "A regular expression that must match for a system interface\nto be included in the AD DC interfaces list.\n"
```
anoopcs9, line 265: nit:

```suggestion
                "description": "A regular expression that must match for a network interface\nto be included in the AD DC interfaces list.\n"
```

and other occurrences.

anoopcs9 (approved):

sambacc/schema/conf-v0.schema.json
phlogistonjohn: Good point, I'll fix that in a follow up change, thanks!

phlogistonjohn: https://github.com/samba-in-kubernetes/sambacc/pull/110

PR #114, opened 2024-04-19
workflows: disable fail-fast option for distro matrix

phlogistonjohn:
There's a consistent failure on the f38 build that causes the other builds to terminate early. Unset the fail-fast option to allow the other jobs to run while we investigate the failure and/or wait for f40.


synarete (approved):
lgtm

gd (approved):
LGTM

PR #117, opened 2024-06-11
sambacc: simplify the closure function writing ctdb config

phlogistonjohn:
A minor cleanup in the ctdb.py module. If we are going to use a closure we might as well go all the way with it.


anoopcs9 (approved):
lgtm, thanks.

synarete (approved):
lgtm

PR #118, opened 2024-06-13
ctdb: modularize the ctdb.py code a little

phlogistonjohn:
Make it possible for future changes to store the metadata differently.

### DIFF
--- a/sambacc/typelets.py
+++ b/sambacc/typelets.py
@@ -20,8 +20,17 @@ remember/redefine.
 """
 
 from types import TracebackType
+import sys
 import typing
 
 ExcType = typing.Optional[typing.Type[BaseException]]
 ExcValue = typing.Optional[BaseException]
 ExcTraceback = typing.Optional[TracebackType]
+
+
+if sys.version_info >= (3, 11):
+    from typing import Self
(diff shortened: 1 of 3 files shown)


sambacc/typelets.py
```diff
@@ -20,8 +20,17 @@
...
 ExcValue = typing.Optional[BaseException]
 ExcTraceback = typing.Optional[TracebackType]
+
+
+if sys.version_info >= (3, 11):
+    from typing import Self
+elif typing.TYPE_CHECKING:
+    from typing_extensions import Self
+else:
+    Self = typing.Any
```
synarete, line 36: Nice idea!

When checking python version, I usually use [platform.python_version_tuple](https://docs.python.org/3/library/platform.html#platform.python_version_tuple), but `sys.version_info` is just fine here.

synarete (approved):
lgtm

phlogistonjohn:
@anoopcs9 PTAL

anoopcs9 (approved):
I was bit reluctant to review this change with all the language specific idioms used for introducing the new classes. But overall I think the rearrangement looks good.

PR #119, opened 2024-06-14
update black version

phlogistonjohn:
Depends on: #118 

Add some docstrings and then update the version of the black formatter tool our tox envs are allowed to use.

PR #120, opened 2024-06-18
commands: add new generic --skip-if command line options 

phlogistonjohn:
Depends on: #119

Previously, sambacc had the `--skip-if-file=` command line option that
would skip running a command if a named file exists. I have a need for
doing something similar for environment variable contents. In order to
avoid needing to add a plethora of new --skip-if-x type options over
time, add a more generic `--skip-if=` option and have the value contain
an initial prefix like `file:` or `env:` that determines how the rest of
the value will be parsed and how the skip will be determined (or not).
Deprecate `--skip-if-file`.

Maybe it's a case of over engineering, but at least it's easy to unit
test.

Examples:
* `--skip-if=env:CHAT==yes`, if the environment variable `CHAT` contains
  a value equal to "yes" the command will be skipped
* `--skip-if=file:/foo/bar` if the file `/foo/bar` exists the command
  will be skipped
* `--skip-if=file:!/foo/bar` if the file `/foo/bar` does not exist the command
  will be skipped
* `--skip-if=env:MODE!=777`, if the environment variable `MODE` contains
  a value not equal to "777" the command will be skipped
* `--skip-if=always:`, always skip this command (mainly to be used for
  testing/hacking on things)

### DIFF
--- a/sambacc/commands/dcmain.py
+++ b/sambacc/commands/dcmain.py
@@ -19,10 +19,10 @@
 import typing
 
 from . import addc
+from . import skips
 from .cli import Fail
 from .main import (
     CommandContext,
-    action_filter,
     enable_logging,
     env_to_cli,
     global_args,
@@ -41,9 +41,10 @@ def main(args: typing.Optional[typing.Sequence[str]] = None) -> None:
         raise Fail("missing container identity")
 
     pre_action(cli)
-    skip = action_filter(cli)
+    ctx = CommandContext(cli)
+    skip = skips.test(ctx)
     if skip:
-        print(f"Action skipped: {skip}")
+        print(f"Command Skipped: {skip}")
         return
     cfunc = getattr(cli, "cfunc", default_cfunc)
     cfunc(CommandContext(cli))
--- a/sambacc/commands/main.py
+++ b/sambacc/commands/main.py
@@ -36,6 +36,7 @@ from . import dns  # noqa: F401
 from . import initialize  # noqa: F401
 from . import join  # noqa: F401
 from . import run  # noqa: F401
+from . import skips
 from . import users  # noqa: F401
 from .cli import commands, Fail, Parser
 
@@ -99,9 +100,22 @@ def global_args(parser: Parser) -> None:
         help="Wrap samba commands within a supplied command prefix",
     )
     parser.add_argument(
+        "--skip-if",
+        dest="skip_conditions",
+        action="append",
+        type=skips.parse,
+        help=(
+            "Skip execution based on a condition. Conditions include"
+            " 'file:<path>', 'env:<var>(==|!=)<value>', and 'always:'."
+            " (Pass `?` for more details)"
+        ),
+    )
+    parser.add_argument(
         "--skip-if-file",
         action="append",
-        help="Perform no action if the specified path exists.",
+        dest="skip_conditions",
+        type=skips.SkipFile.parse,
+        help="(DEPRECATED) Perform no action if the specified path exists.",
     )
     parser.add_argument(
         "--validate-config",
@@ -309,13 +323,6 @@ def enable_logging(cli: typing.Any) -> None:
     logger.addHandler(handler)
 
 
-def action_filter(cli: typing.Any) -> typing.Optional[str]:
-    for path in cli.skip_if_file or []:
-        if os.path.exists(path):
-            return f"skip-if-file: {path} exists"
-    return None
-
-
 def main(args: typing.Optional[typing.Sequence[str]] = None) -> None:
     cli = commands.assemble(arg_func=global_args).parse_args(args)
     env_to_cli(cli)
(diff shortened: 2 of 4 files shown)


phlogistonjohn:
cc: @avanthakkar - for an additional review & peek into this code base

sambacc/commands/dcmain.py
```diff
@@ -41,9 +41,10 @@ def main(args: typing.Optional[typing.Sequence[str]] = None) -> None:
         raise Fail("missing container identity")
 
     pre_action(cli)
-    skip = action_filter(cli)
+    ctx = CommandContext(cli)
+    skip = skips.test(ctx)
     if skip:
-        print(f"Action skipped: {skip}")
+        print(f"Command Skipped: {skip}")
```
synarete, line 47: Should probably be "Command skipped" (lower case on 2nd word)

sambacc/commands/skips.py
```diff
@@ -0,0 +1,188 @@
...
+        if not value:
+            raise ValueError("missing path")
+        if value[0] == "!":
+            obj.inverted = True
+            value = value[1:]
+        obj.path = value
+        return obj
+
+    def test(self, ctx: Context) -> Optional[str]:
+        exists = os.path.exists(self.path)
```
synarete, line 73: Why not use modern `pathlib.Path` ?

synarete (approved):
lgtm, but see minor comments

sambacc/commands/dcmain.py
phlogistonjohn: I intentionally wanted that one to stand out and use title case.

sambacc/commands/skips.py
phlogistonjohn: A mix of force of habit and that the code was already using os.path.

phlogistonjohn:
I noticed a stray debugging print statement was still in the unit test file. It's now removed.

synarete (dismissed):

sambacc/commands/skips.py
```diff
@@ -0,0 +1,188 @@
...
+        elif cls._NEQ in value:
+            op = cls._NEQ
+        else:
+            raise ValueError("invalid SkipEnv: missing or invalid operation")
+        lhv, rhv = value.split(op, 1)
+        return cls(op, lhv, rhv)
+
+    def test(self, ctx: Context) -> Optional[str]:
+        env_val = os.environ.get(self.var_name)
+        if self.op == self._EQ and env_val == self.target_value:
```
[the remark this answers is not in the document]
phlogistonjohn: I prefer having the unique reason strings , but thanks for the suggestion.

sambacc/commands/skips.py
```diff
@@ -0,0 +1,188 @@
...
+            op = cls._EQ
+        elif cls._NEQ in value:
+            op = cls._NEQ
+        else:
+            raise ValueError("invalid SkipEnv: missing or invalid operation")
+        lhv, rhv = value.split(op, 1)
+        return cls(op, lhv, rhv)
+
+    def test(self, ctx: Context) -> Optional[str]:
+        env_val = os.environ.get(self.var_name)
```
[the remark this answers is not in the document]
phlogistonjohn: I prefer os.environ as it feels more pythonic to have a "special" dict and use typical dict methods rather than relying on specific calls unique to the environment variable handling like C.

sambacc/commands/skips.py
```diff
@@ -0,0 +1,188 @@
...
+    prefixed with the skip "type" (the skip type's NAME).
+    """
+    if value == "?":
+        # A hack to avoid putting tons of documentation into the help output.
+        raise argparse.ArgumentTypeError(_help_info())
+    for sk in _SKIP_TYPES:
+        assert issubclass(sk, SkipIf)
+        prefix = f"{sk.NAME}:"
+        plen = len(prefix)
+        if not value.startswith(prefix):
```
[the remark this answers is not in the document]
phlogistonjohn: Good call. Updated.

sambacc/commands/skips.py
```diff
@@ -0,0 +1,188 @@
...
+        raise NotImplementedError()  # pragma: nocover
+
+    @classmethod
+    def parse(cls, value: str) -> Self:
+        """Parse a string into a skip class arguments."""
+        raise NotImplementedError()  # pragma: nocover
+
+
+class SkipFile(SkipIf):
+    """Skip execution if a file is, or is not, present. For example,
```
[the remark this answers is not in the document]
phlogistonjohn: Updated.

avanthakkar (approved):
LGTM!

synarete (approved):
lgtm

Shwetha-Acharya (approved):

sambacc/commands/main.py
```diff
@@ -98,10 +99,23 @@ def global_args(parser: Parser) -> None:
...
         help="Wrap samba commands within a supplied command prefix",
     )
+    parser.add_argument(
+        "--skip-if",
+        dest="skip_conditions",
+        action="append",
+        type=skips.parse,
+        help=(
+            "Skip execution based on a condition. Conditions include"
+            " 'file:<path>', 'env:<var>(==|!=)<value>', and 'always:'."
```
anoopcs9, line 109: I know that we have `?` for more detailed explanation but shouldn't we mention about `file:!<path>`  also here?

phlogistonjohn: I'll add it as " 'file:[!]<path>', ... in a follow up PR. Sound OK?

anoopcs9 (approved):

PR #122, opened 2024-06-24
commands: prepare ctdb commands for using cluster meta objects

phlogistonjohn:
Preparatory work for being able to store this cluster metadata in rados.

### DIFF
--- a/sambacc/commands/ctdb.py
+++ b/sambacc/commands/ctdb.py
@@ -23,6 +23,8 @@ import socket
 import typing
 
 from sambacc import ctdb
+from sambacc import jfile
+from sambacc.simple_waiter import Waiter
 
 from .cli import best_waiter, commands, Context, Fail
 
@@ -86,6 +88,13 @@ def _ctdb_general_node_args(parser: argparse.ArgumentParser) -> None:
         "--persistent-path",
         help="Path to a persistent path for storing nodes file",
     )
+    parser.add_argument(
+        "--metadata-source",
+        help=(
+            "Specify location of cluster metadata state-tracking object."
+            " This can be a file path or a URI-style identifier."
+        ),
+    )
 
 
 def _ctdb_set_node_args(parser: argparse.ArgumentParser) -> None:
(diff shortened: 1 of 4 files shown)


phlogistonjohn:
cc @avanthakkar

synarete (dismissed):
lgtm

sambacc/commands/ctdb.py
```diff
@@ -86,6 +88,13 @@ def _ctdb_general_node_args(parser: argparse.ArgumentParser) -> None:
         "--persistent-path",
         help="Path to a persistent path for storing nodes file",
     )
+    parser.add_argument(
+        "--metadata-state",
```
anoopcs9, line 92: I think `--metadata-state` doesn't fully match the intended type of value which basically points to an endpoint as source of truth. Isn't it better to use `--metadata-source` or similar?

sambacc/commands/ctdb.py
```diff
@@ -156,6 +173,49 @@ def identity(self):
...
+    @property
+    def cluster_meta_uri(self) -> str:
+        """Return a cluster meta uri value."""
+        values = (
+            # cli takes highest precedence
+            self._ctx.cli.metadata_state,
+            # _nodes_json should only be set if user set it using the old key
+            self._nodes_json,
+            # default or customized value on current key
+            self._cluster_meta_uri,
```
anoopcs9, line 185: May be reverse the order here to prefer new method?

sambacc/commands/ctdb.py
phlogistonjohn: Sure. Will update.

sambacc/ctdb.py
```diff
@@ -338,9 +338,23 @@ def manage_nodes(
     pause_func: typing.Callable,
```
anoopcs9, line 338: `manage_nodes()` now exists for testing purpose only, right?

anoopcs9 (approved):
See below for a small remark.

Otherwise lgtm, thanks.

synarete (approved):
lgtm

sambacc/ctdb.py
phlogistonjohn: Yeah, there are now a bunch of functions that are consumed only by the test suite. It's a TODO item to circle back and update the tests and eventually remove the functions.

PR #123, opened 2024-06-24
Extend support for ctdb cluster meta stored in ceph rados

phlogistonjohn:
Depends on #122

Add support to rados module to implement the ClusterMeta protocol.
Add rados support to sambacc ctdb commands.
Add a helpful CLI option for getting node number from env (it's a bit unrelated to the rados focus of the other changes but I didn't want to file a whole PR just for that).

### DIFF
--- a/sambacc/rados_opener.py
+++ b/sambacc/rados_opener.py
@@ -388,15 +388,17 @@ def is_rados_uri(uri: str) -> bool:
     return uri.startswith("rados:")
 
 
-def enable_rados_url_opener(
+def enable_rados(
     cls: typing.Type[url_opener.URLOpener],
     *,
     client_name: str = "",
     full_name: bool = False,
 ) -> None:
-    """Extend the URLOpener type to support pseudo-URLs for rados
-    object storage. If rados libraries are not found the function
-    does nothing.
+    """Enable Ceph RADOS support in sambacc.
+    As as side-effect it will extend the URLOpener type to support pseudo-URLs
+    for rados object storage. It will also enable the
+    ClusterMetaRADOSObject.create_from_uri constructor. If rados libraries are
+    not found the function does nothing.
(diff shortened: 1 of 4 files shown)


phlogistonjohn:
cc @avanthakkar

synarete (approved):
lgtm

sambacc/rados_opener.py
```diff
@@ -198,6 +251,38 @@ def write(self, s: typing.Any) -> int:
...
+        self._ioctx.lock_exclusive(
+            self._key,
+            name,
+            cookie,
+            desc=self._lock_description,
+            duration=self._lock_duration,
+        )
+
+    def _acquire_lock_exclusive(
+        self, name: str, cookie: str, *, delay: int = 1
```
avanthakkar, line 268: Maybe not too crucial but can be limit retries here instead of  trying indefinitely?
I think if the condition that causes the `ObjectBusy` exception persists indefinitely it could lead to a hung process that never completes.

```suggestion
        self, name: str, cookie: str, *, delay: int = 1, max_retries: int = 10
```

phlogistonjohn: Fair enough. However, I'd prefer to handle that in a follow up PR.

avanthakkar: Sounds good to me!

avanthakkar (approved):

anoopcs9 (changes requested):
The very last commit stands out. Shall we save `commands: support acquiring a node number from the environment ` for a later PR which could be more related?

anoopcs9 (approved):
> The very last commit stands out. Shall we save `commands: support acquiring a node number from the environment ` for a later PR which could be more related?

Ah.. I should have looked at PR description in detail 😉 .

PR #124, opened 2024-06-29
Add some more ctdb related hacks

phlogistonjohn:
Add some additional options to the ctdb related sambacc commmands. The --write-node option in particular is a bit hacky but it helps me get on with prototyping samba+ctdb in containers under ceph.

### DIFF
--- a/sambacc/ctdb.py
+++ b/sambacc/ctdb.py
@@ -586,6 +586,23 @@ def _convert_tdb_file(tdb_path: str, dest_dir: str, pnn: int = 0) -> None:
     subprocess.check_call(list(cmd))
 
 
+def archive_tdb(iconfig: config.InstanceConfig, dest_dir: str) -> None:
+    """Arhive TDB files into a given directory."""
+    # TODO: these paths should be based on our instance config, not hard coded
+    try:
+        os.mkdir(dest_dir)
+        _logger.debug("dest_dir: %r created", dest_dir)
+    except FileExistsError:
+        _logger.debug("dest_dir: %r already exists", dest_dir)
+    for tdbfile in _SRC_TDB_FILES:
+        for parent in _SRC_TDB_DIRS:
+            tdb_path = os.path.join(parent, tdbfile)
+            if _has_tdb_file(tdb_path):
+                dest_path = os.path.join(dest_dir, tdbfile)
+                _logger.info("archiving: %r -> %r", tdb_path, dest_path)
+                os.rename(tdb_path, dest_path)
+
+
 def check_nodestatus(cmd: samba_cmds.SambaCommand = samba_cmds.ctdb) -> None:
     cmd_ctdb_check = cmd["nodestatus"]
     samba_cmds.execute(cmd_ctdb_check)
--- a/sambacc/ctdb.py
+++ b/sambacc/ctdb.py
@@ -483,6 +483,26 @@ def _node_update(cmeta: ClusterMeta, real_path: str) -> bool:
     return True
 
 
+def cluster_meta_to_nodes(cmeta: ClusterMeta, real_path: str) -> None:
+    """Write a nodes file based on the current content of the cluster
+    metadata."""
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
cc  @avanthakkar

sambacc/ctdb.py
```diff
@@ -483,6 +483,26 @@ def _node_update(cmeta: ClusterMeta, real_path: str) -> bool:
...
+        ctdb_nodes: list[str] = [""] * pnn_max
+        for entry in nodes:
+            pnn = entry["pnn"]
+            expected_line = _entry_to_node(ctdb_nodes, entry)
+            ctdb_nodes[pnn] = expected_line
+        _logger.info("Will write nodes: %s", ctdb_nodes)
+        with open(real_path, "w") as nffh:
+            write_nodes_file(nffh, ctdb_nodes)
+            nffh.flush()
+            os.fsync(nffh)
```
synarete, line 503: Could you please remind me why we need this pattern of doing both `flush` and `os.fsync` ? 
It is repeated in this module; maybe have a local helper-function with minimal doc-string.

sambacc/ctdb.py
```diff
@@ -586,6 +586,23 @@ def _convert_tdb_file(tdb_path: str, dest_dir: str, pnn: int = 0) -> None:
...
+    """Arhive TDB files into a given directory."""
+    # TODO: these paths should be based on our instance config, not hard coded
+    try:
+        os.mkdir(dest_dir)
+        _logger.debug("dest_dir: %r created", dest_dir)
+    except FileExistsError:
+        _logger.debug("dest_dir: %r already exists", dest_dir)
+    for tdbfile in _SRC_TDB_FILES:
+        for parent in _SRC_TDB_DIRS:
+            tdb_path = os.path.join(parent, tdbfile)
```
synarete, line 599: Instead of double nested `for` loop, consider something like:

```
from itertools import product
from pathlib import Path

for tdb_path in [ Path(p) / f for p, f in product(_SRC_TDB_DIRS, _SRC_TDB_FILES) ]:
    ...
```

synarete (commented):
Looks fine, minor style comments

sambacc/ctdb.py
phlogistonjohn: This is how the older, migrate functions is written. I don't want to deviate from the pattern now.

sambacc/ctdb.py
phlogistonjohn: To be fair it is a copy-paste that was tweaked. The goal of the pattern is to call `flush` to make sure any buffers internal to the process get flushed to the OS. Then `fsync` tells the os to flush any buffers it has and ensure the entire file is written to the storage device. Normally, flush is not required as it is done also by close, but to use fync we need an fd - so we end up having to manually flush first.

sambacc/ctdb.py
synarete: Fair enough

phlogistonjohn:
@synarete these answers good enough? && ping @anoopcs9

sambacc/ctdb.py
synarete: Fair enough. I still think it would be nice to have it local helper-function but this can wait to future cleanups PR.

synarete (approved):
lgtm

sambacc/ctdb.py
phlogistonjohn: Yeah, if you note I'm not writing very many tests for some of this stuff either. Normally, I'd criticize myself for that but I'm doing a bunch of prototyping simultaneously on the ceph codebase and sambacc. I'm trying to get as many of these into the codebase here in sambacc so I don't end up with massive private branches on both sides. :-)
I'd welcome anyone to jump in and help contribute tests and cleanups on top. :-)

phlogistonjohn:
@anoopcs9 2nd review please?
I now have #126 that depends on this PR as well.

anoopcs9 (approved):

PR #127, opened 2024-07-15
ctdb: add ctdb-monitor-nodes command 

phlogistonjohn:
Add a `ctdb-monitor-nodes` command - this command will monitor the
cluster meta and update the containers/ctdb when/if it changes.
The `--write-nodes` parameter is used to instruct the command to
write the ctdb nodes file on changes. The `--reload` option is used to
control what "nodes" issue `ctdb reloadnodes` commands: leader (only the
cluster leader), all (all nodes that see the cluster meta change), and never
(never issue reloadnodes commands).


sambacc/ctdb.py
```diff
@@ -490,17 +488,96 @@ def cluster_meta_to_nodes(cmeta: ClusterMeta, real_path: str) -> None:
...
+        _save_nodes(real_path, ctdb_nodes)
+
+
+def _cluster_meta_to_ctdb_nodes(nodes: list[dict]) -> list[str]:
+    pnn_max = max(n["pnn"] for n in nodes) + 1  # pnn is zero indexed
+    ctdb_nodes: list[str] = [""] * pnn_max
+    for entry in nodes:
+        pnn = entry["pnn"]
+        expected_line = _entry_to_node(ctdb_nodes, entry)
+        ctdb_nodes[pnn] = expected_line
```
synarete, line 502: Consider making it a one-liner:
`ctdb_nodes[pnn] = _entry_to_node(ctdb_modes,  entry)`

sambacc/ctdb.py
```diff
@@ -490,17 +488,96 @@ def cluster_meta_to_nodes(cmeta: ClusterMeta, real_path: str) -> None:
...
+            _logger.warning("no leader locator: not calling reloadnodes")
+        else:
+            # use the leader locator to only issue the reloadnodes command once
+            # for a change instead of all the nodes "spamming" the cluster
+            with leader_locator as ll:
+                if ll.is_leader():
+                    _logger.info("running: ctdb reloadnodes")
+                    subprocess.check_call(list(samba_cmds.ctdb["reloadnodes"]))
+                else:
+                    _logger.info("node is not leader. skipping reloadnodes")
```
synarete, line 580: IMHO such long logic deserves to be split into one (ore more) separate function(s). It makes the life of future reader/maintainer of this code easier.

sambacc/ctdb.py
phlogistonjohn: ok, done in a separate commit to preserve the move-only nature of the original commit.

sambacc/ctdb.py
phlogistonjohn: done.
If you think this is a long function I dare you to read the cephadm mgr module code.

synarete: I doing so right now... sigh...

synarete (approved):
lgtm

phlogistonjohn:
@anoopcs9 ping

PR #128, opened 2024-07-18
ctdb: nodes list command, wait for ctdb option

phlogistonjohn:
Add a `samba-container ctdb-list-nodes` command that emits a "nodes file" content to the stdout of the process.
Add hooks to set up a ctdb "nodes_cmd" config file option that tells ctdb to execute a command to fetch the nodes rather than read a file.

Also add a `--wait-for=ctdb` option to avoid starting certain services before ctdb is available and causing unnecessary container restarts.

### DIFF
--- a/sambacc/commands/run.py
+++ b/sambacc/commands/run.py
@@ -16,8 +16,11 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>
 #
 
+import contextlib
 import logging
+import signal
 import time
+import typing
 
 from sambacc import samba_cmds
 import sambacc.paths as paths
@@ -103,6 +106,21 @@ def _run_container_args(parser):
     )
 
 
+_COND_TIMEOUT = 5 * 60
+
+
+@contextlib.contextmanager
+def _timeout(timeout: int) -> typing.Iterator[None]:
+    def _handler(sig: int, frame: typing.Any) -> None:
+        raise RuntimeError("timed out waiting for conditions")
+
+    signal.signal(signal.SIGALRM, _handler)
+    signal.alarm(timeout)
+    yield
+    signal.alarm(0)
+    signal.signal(signal.SIGALRM, signal.SIG_DFL)
+
+
 @commands.command(name="run", arg_func=_run_container_args)
 def run_container(ctx: Context) -> None:
     """Run a specified server process."""
@@ -110,9 +128,10 @@ def run_container(ctx: Context) -> None:
         raise Fail("can not specify both --no-init and --setup")
 
     if ctx.cli.wait_for:
-        conditions = [_wait_for_conditions[n]() for n in ctx.cli.wait_for]
-        while not all(c.met(ctx) for c in conditions):
-            time.sleep(1)
+        with _timeout(_COND_TIMEOUT):
+            conditions = [_wait_for_conditions[n]() for n in ctx.cli.wait_for]
+            while not all(c.met(ctx) for c in conditions):
+                time.sleep(1)
 
     # running servers expect to make use of ctdb whenever it is configured
     ctx.expects_ctdb = True
--- a/sambacc/commands/run.py
+++ b/sambacc/commands/run.py
@@ -16,6 +16,9 @@
 # along with this program.  If not, see <http://www.gnu.org/licenses/>
 #
 
+import logging
+import time
+
 from sambacc import samba_cmds
 import sambacc.paths as paths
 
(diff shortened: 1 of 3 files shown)


sambacc/ctdb.py
```diff
@@ -200,6 +203,19 @@ def ensure_ctdb_node_present(
...
+    ctdb_params = iconfig.ctdb_config()
+    if not ctdb_params.get("nodes_cmd"):
+        return
+    cmd = ctdb_params["nodes_cmd"]
+    with open(cmd, "w") as fh:
+        fh.write("#!/bin/sh\n")
+        fh.write("exec samba-container ctdb-list-nodes\n")
+        fh.flush()
+        os.fchmod(fh.fileno(), 0o755)
+        os.fsync(fh)
```
anoopcs9, line 216: Why are we creating another executable with the required command? Why don't we directly invoke `ctdb-list-nodes` command to _samba-container_? Or is this done to not to interfere with current execution environment's `STDOUT`?

phlogistonjohn: Right now, the patches to CTDB only support a single command name, no arguments. It was simpler for me to work around that on the sambacc side for now than it was to figure out the right way to do it in C. Help appreciated on that side: https://gitlab.com/phlogistonjohn/samba/-/commit/3d6efac6e8ae1056149fb14ada40f6ff5db0f7c6#61bd70ccace1b02c4b6dfa6bd251aadbd56de122_167_188

phlogistonjohn: Resolved in samba repo.

sambacc/ctdb.py
```diff
@@ -704,6 +704,42 @@ def check_nodestatus(cmd: samba_cmds.SambaCommand = samba_cmds.ctdb) -> None:
     samba_cmds.execute(cmd_ctdb_check)
 
 
+def _read_command_pnn(cmd: samba_cmds.SambaCommand) -> typing.Optional[int]:
+    """Run a ctdb command assuming it returns a pnn value. Return the pnn as an
+    int on success, None on command failure.
+    """
+    try:
+        out = subprocess.check_output(list(cmd))
+        pnntxt = out.decode("utf8").strip()
```
synarete, line 713: I think this could yield `UnicodeDecodeError`. Should be handled with proper `except`.

sambacc/ctdb.py
```diff
@@ -704,6 +704,42 @@ def check_nodestatus(cmd: samba_cmds.SambaCommand = samba_cmds.ctdb) -> None:
     samba_cmds.execute(cmd_ctdb_check)
 
 
+def _read_command_pnn(cmd: samba_cmds.SambaCommand) -> typing.Optional[int]:
```
synarete, line 707: I am not a big fan of this mixed int-or-None return value. I would prefer a more explicit typing; something like `tuple[int, bool]`. But I guess its just a matter of style.

sambacc/commands/run.py
```diff
@@ -67,6 +108,12 @@ def run_container(ctx: Context) -> None:
     """Run a specified server process."""
     if ctx.cli.no_init and ctx.cli.setup:
         raise Fail("can not specify both --no-init and --setup")
+
+    if ctx.cli.wait_for:
+        conditions = [_wait_for_conditions[n]() for n in ctx.cli.wait_for]
+        while not all(c.met(ctx) for c in conditions):
+            time.sleep(1)
+
```
synarete, line 116: This has the potential of forever loop. Better limit the number of iterations/time.

synarete (commented):
Looks fine. Few minor comments.

sambacc/commands/run.py
```diff
@@ -47,6 +77,17 @@ def _run_container_args(parser):
             " The special 'init-all' name will perform all known setup steps."
         ),
     )
+    _wait_for_choices = _wait_for_conditions.keys()
+    parser.add_argument(
+        "--wait-for",
+        action="append",
+        choices=_wait_for_choices,
+        help=(
+            "Specify a condition to wait for prior to starting the server"
```
anoopcs9, line 86: FYI: `ctdb event script enable legacy 50.samba`

CTDB can manage Samba services(_nmb_ for `nmbd`, _smb_ for `smbd`, _winbind_ for `winbindd` etc) with the help of event scripts as mentioned in the [docs](https://ctdb.samba.org/manpages/ctdb-script.options.5.html) under "SERVICE CONFIGURATION". This is in fact currently in use by sit-environment to setup [clustered samba](https://github.com/samba-in-kubernetes/sit-environment/blob/main/playbooks/ansible/roles/ctdb.setup/tasks/main.yml#L96).

phlogistonjohn: I know, but we use containers and this script is not appropriate for that environment, it uses the `service` command. This is one of the things I complain about when I say that ctdb was not designed for containers and interoperating with a higher level of orchestration. IMO, most of the event scripts are not appropriate for containers.

In ceph orch I was looking at two choices... either make the primary container variable (switch the primary between ctdb and smbd) depending on deployment type or always having smbd be the primary and then making it delay the start. I went with the latter for consistentcy when debugging and thinking about the system but I needed a slightly cleaner way for waiting to ctdb to be ready than just delaying from a fixed length of time.

sambacc/ctdb.py
phlogistonjohn: Yeah, sorry but I find your suggestion less idiomatic. Hope you don't mind.

sambacc/ctdb.py
phlogistonjohn: It's extremely unlikely given the output of the ctdb command. I'd prefer the let the exception bubble up to the top level if it occurs.

sambacc/commands/run.py
phlogistonjohn: That was intentional. But I suppose having a very long timeout is ok because something like 5 minutes would clearly be outside the expected bounds of startup and provide some diagnostic information in the failure.
I'll make this change.

phlogistonjohn: done. ptal.

sambacc/commands/run.py
anoopcs9: I completely get it, thanks.

sambacc/ctdb.py
synarete: Fair enough

synarete (approved):
lgtm

anoopcs9 (approved):

PR #130, opened 2024-08-16
sambacc: add a retry loop to ctdb.monitor_cluster_meta_changes

phlogistonjohn:
Add a loop that tries the `ctdb reloadnodes` command after an increasing delay. This is an attempt to fix a condition where ctdbd is apparently not ready to handle the `ctdb reloadnodes` command. In this case the command would be run, but fail and an exception would be raised in the monitor_cluster_meta_changes function would raise an exception. This would be caught by the command-level retry loop. However, this command-level retry loop will simply re-run monitor_cluster_meta_changes and this function now no longer has the same initial clustermeta state and has effectively "forgotten" that it needs to run reloadnodes.  This new retry loop adds a level of error handling inside the monitor_cluster_meta_changes function so that we will retry with a bounded number of attempts.

### DIFF
--- a/sambacc/ctdb.py
+++ b/sambacc/ctdb.py
@@ -20,6 +20,7 @@ import enum
 import logging
 import os
 import subprocess
+import time
 import typing
 
 from sambacc import config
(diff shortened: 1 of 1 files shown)


sambacc/ctdb.py
```diff
@@ -572,7 +573,23 @@ def monitor_cluster_meta_changes(
...
+
+
+def _maybe_reload_nodes_retry(
+    leader_locator: typing.Optional[leader.LeaderLocator] = None,
+    reload_all: bool = False,
+    *,
+    tries: int = 5,
+) -> None:
+    for idx in range(tries):
+        time.sleep(1 << idx)
```
synarete, line 586: Minor: have upper bound to sleep time. Something like `time.sleep(min(1 << idx, 60))`. Just in case someone naively increases `tries` to large value.

synarete (approved):
LGTM. See minor comment.

anoopcs9 (approved):
lgtm.

PR #134, opened 2024-08-27
sambacc: fix constructing interface list for ctdb public_addresses

phlogistonjohn:
The samba wiki and documentation clearly state that the list of interfaces should be comma separated but I incorrectly joined them with spaces. Fix this and one other small suboptimal line in the same function.

### DIFF
--- a/sambacc/ctdb.py
+++ b/sambacc/ctdb.py
@@ -197,9 +197,9 @@ def _write_public_addresses_file(
     fh: typing.IO, addrs: list[PublicAddrAssignment]
 ) -> None:
     for entry in addrs:
-        fh.write(f"{entry['address']}")
+        fh.write(entry["address"])
         if entry["interfaces"]:
-            ifaces = " ".join(entry["interfaces"])
+            ifaces = ",".join(entry["interfaces"])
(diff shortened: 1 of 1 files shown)


xhernandez (approved):

anoopcs9 (approved):
> The samba wiki and documentation clearly state that the list of interfaces should be comma separated but I incorrectly joined them with spaces. Fix this and one other small suboptimal line in the same function.

As a follow up to my previous https://github.com/samba-in-kubernetes/sambacc/pull/133#discussion_r1732733947, I wasn't aware of such a configuration with multiple interfaces until I thoroughly looked into man [ctdb(7)](https://ctdb.samba.org/manpages/ctdb.7.html)(following @phlogistonjohn 's suggestion) to find just one example and that too under a specially catered "POLICY ROUTING" section. Judging from the details this is most likely an advanced network configuration which might have to be clubbed with some additional event scripts. But for now let's keep this option to specify multiple comma(`,`) separated interfaces for each network address.

Anyway it is a good find.

obnoxxx (approved):
LGTM

Issue #136, opened 2024-09-02
Improve hostname lookup for ctdb nodes


avanthakkar:
The hostname [lookup](https://github.com/samba-in-kubernetes/sambacc/blob/master/sambacc/commands/ctdb.py#L264) for ctdb nodes  is limited to IPv4 and uses a loopback filtering hack. 
It should be improved to support multi-homed systems, mixed IPv4/IPv6 environments, and potentially use `socket.getaddrinfo()` for better address handling.

PR #137, opened 2024-09-02
ctdb: improve hostname lookup for ctdb nodes

avanthakkar:
Updated the hostname lookup function to support multi-homed systems and mixed IPv4/IPv6 environments by using `socket.getaddrinfo`. 
This change removes the previous loopback filtering hack, now explicitly filtering loopback addresses and prioritizing IPv4 addresses where available. Improved error handling for cases with no valid IP addresses found.

Fixes: https://github.com/samba-in-kubernetes/sambacc/issues/136

### DIFF
--- a/sambacc/commands/ctdb.py
+++ b/sambacc/commands/ctdb.py
@@ -261,12 +261,39 @@ def ctdb_migrate(ctx: Context) -> None:
         ctdb.archive_tdb(ctx.instance_config, ctx.cli.archive)
 
 
-def _lookup_hostname(hostname):
-    # XXX this is a nasty little hack.
-    ips = socket.gethostbyname_ex(hostname)[2]
-    addr = [ip for ip in ips if ip != "127.0.0.1"][0]
-    _logger.info(f"Determined address for {hostname}: {addr}")
-    return addr
+def _lookup_hostname(hostname: str) -> str:
+    try:
+        addrinfo = socket.getaddrinfo(
+            hostname,
+            None,
+            family=socket.AF_UNSPEC,
+            type=socket.SOCK_STREAM,
+        )
+        ipv6_address = None
+
+        for entry in addrinfo:
+            family, _, _, _, sockaddr = entry
+            ip_address = sockaddr[0]
+
+            if ip_address.startswith("127.") or ip_address == "::1":
+                continue
+
+            if family == socket.AF_INET:
+                return ip_address
+
+            if family == socket.AF_INET6 and ipv6_address is None:
+                ipv6_address = ip_address
+
+        if ipv6_address:
+            return ipv6_address
+
+        raise RuntimeError(
+            f"No valid IP address found for hostname '{hostname}'."
+        )
+
+    except socket.gaierror as e:
+        _logger.error(f"Failed to resolve hostname '{hostname}': {e}")
+        raise
 
 
 @commands.command(name="ctdb-set-node", arg_func=_ctdb_set_node_args)


sambacc/commands/ctdb.py
```diff
@@ -261,12 +261,48 @@ def ctdb_migrate(ctx: Context) -> None:
...
+
+        for entry in addrinfo:
+            family, _, _, _, sockaddr = entry
+            ip_address = sockaddr[0]
+
+            if ip_address.startswith("127.") or ip_address == "::1":
+                continue
+
+            if family == socket.AF_INET:
+                ipv4_addresses.append(ip_address)
```
xhernandez, line 280: We are going to return the first IPv4 address if it exists, so there's no need to continue checking other addresses. We can directly return the current address from here.

sambacc/commands/ctdb.py
```diff
@@ -261,12 +261,48 @@ def ctdb_migrate(ctx: Context) -> None:
...
-    addr = [ip for ip in ips if ip != "127.0.0.1"][0]
-    _logger.info(f"Determined address for {hostname}: {addr}")
-    return addr
+def _lookup_hostname(hostname: str) -> str:
+    try:
+        addrinfo = socket.getaddrinfo(
+            hostname, None, family=socket.AF_UNSPEC, type=socket.SOCK_STREAM
+            )
+        ipv4_addresses = []
+        ipv6_addresses = []
```
xhernandez, line 270: We are only interested in the first address. There's no need to keep a list.

sambacc/commands/ctdb.py
```diff
@@ -261,12 +261,48 @@ def ctdb_migrate(ctx: Context) -> None:
...
+        else:
+            raise ValueError(
+                f"No valid IP address found for hostname '{hostname}'."
+                )
+
+    except socket.gaierror as e:
+        _logger.error(f"Failed to resolve hostname '{hostname}': {e}")
+        raise
+
+    except IndexError:
```
xhernandez, line 302: Is this to handle errors from line 274 ? if so I think we shouldn't abort the search, just ignore that entry and continue with the next one.

sambacc/commands/ctdb.py
```diff
@@ -261,12 +261,48 @@ def ctdb_migrate(ctx: Context) -> None:
...
+            ipv6_addresses[0] if ipv6_addresses else None
+            )
+
+        if preferred_address:
+            _logger.info(
+                f"Determined address for {hostname}: {preferred_address}"
+                )
+            return preferred_address
+        else:
+            raise ValueError(
```
xhernandez, line 294: I'm not sure if `ValueError` is the most appropriate exception here. It's normally used for invalid values passed to the function. In this case it's just that something that the function tried to get is not found. Maybe `RuntimeError` could be more appropriate.

sambacc/commands/ctdb.py
avanthakkar: yeah that makes sense to avoid keeping track of list!

sambacc/commands/ctdb.py
avanthakkar: I see! I think once I remove keeeping track of list it'll get fixed since I'll directly return first address whether it's ipv4 or piv6

avanthakkar: I see! I think once I remove keeeping track of list it'll get fixed since I'll directly return first address whether it's ipv4 or ipv6

sambacc/commands/ctdb.py
avanthakkar: Yes, that's what I tried, but yh makes sense to skip if not found, thanks!

sambacc/commands/ctdb.py
avanthakkar: Fixed!

xhernandez (approved):

Shwetha-Acharya (commented):
LGTM

obnoxxx (approved):
LGTM

PR #139, opened 2024-10-04
ctdb: Leave CTDB log levels at their original defaults

anoopcs9:
There is too much information with the `log level` explicitily set to DEBUG for _ctdbd_ flooding the log file destination. Similarly `script log level` option might also spit out more when set to DEBUG. Instead of removing the options altogether we change the sambacc defaults from DEBUG to NOTCE and ERROR respectively for `log level` and `script log level` options as specified in man [ctdb.conf(5)](https://ctdb.samba.org/manpages/ctdb.conf.5.html).


phlogistonjohn (approved):
lgtm, thanks

avanthakkar (approved):
LGTM

PR #140, opened 2024-11-20
workflows: update test matrix to use fedora 41

phlogistonjohn:
Fedora 41 has been out for some weeks now. Update the test matrix to remove Fedora 39 and use fedora 41.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -7,7 +7,7 @@ queue_rules:
   - name: default
     conditions:
       - check-success=check-commits
-      - check-success=test (fedora-39)
+      - check-success=test (fedora-41)
       - check-success=test (fedora-40)
       - check-success=test (centos-stream9)
       - check-success=dpulls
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
Reminder: this will probably need manual merging as the test names are changing. I'll do it once I have sufficient reviews.

I wonder if I should rename the fedora test jobs something like `fedora-previous` and `fedora-current` or something to avoid this (small bit of) churn? Any opinions?

anoopcs9 (commented):
> I wonder if I should rename the fedora test jobs something like `fedora-previous` and `fedora-current` or something to avoid this (small bit of) churn? Any opinions?

Well, we can't get rid of changes unless we do something as follows:
```
latest=$(curl -s -L https://fedoraproject.org/releases.json | jq -r '[.[].version | select("^[0-9]+$")] | max')

previous=$(( $latest - 1 ))
```

And of course I prefer "latest" over "current".

phlogistonjohn:
@mergifyio rebase

obnoxxx (approved):
LGTM

phlogistonjohn:
> Would you like to go for it with the current PR?

No, I would like to make that change in a different pr. All this slipped off my radar so I would rather fix the immediate issue and then do future proofing later.

PR #143, opened 2024-12-19
ctdb: handle nodes with "GONE" state when listing CTDB nodes

avanthakkar:
Added support for the GONE state in _entry_to_node to exclude nodes no longer part of the cluster.


phlogistonjohn (approved):

synarete (approved):
lgtm

PR #146, opened 2025-02-12
commands: fix a mypy failure in the ctdb commands module

phlogistonjohn:
It seems that a recent mypy release is now determining that certain family types (not AF_INET, or AF_INET6) may return an address tupe with the first element type being an int rather than a str. The (expanded) revealed type is:
```
tuple[
    socket.AddressFamily,
    socket.SocketKind,
    builtins.int,
    builtins.str,
    Union[
        tuple[builtins.str, builtins.int],
        tuple[builtins.str, builtins.int, builtins.int, builtins.int],
        tuple[builtins.int, builtins.bytes]
    ]
]
```

This change ensures that we only check the family values we care about and use a type assertion for those families where we expect the address value to be a str.

FWIW I could not find any documentation for what families have an integer in the first element of the addr tuple. Looking at the python sources it seems like it could be AF_NETLINK, AF_QIPCRTR, AF_VSOCK - all families I have never seen in the wild before :-)

### DIFF
--- a/sambacc/commands/ctdb.py
+++ b/sambacc/commands/ctdb.py
@@ -271,17 +271,19 @@ def _lookup_hostname(hostname: str) -> str:
         )
         ipv6_address = None
 
-        for entry in addrinfo:
-            family, _, _, _, sockaddr = entry
-            ip_address = sockaddr[0]
-
-            if ip_address.startswith("127.") or ip_address == "::1":
-                continue
-
+        for family, _, _, _, sockaddr in addrinfo:
             if family == socket.AF_INET:
+                ip_address = sockaddr[0]
+                assert isinstance(ip_address, str)
+                if ip_address.startswith("127."):
+                    continue
                 return ip_address
 
             if family == socket.AF_INET6 and ipv6_address is None:
+                ip_address = sockaddr[0]
+                assert isinstance(ip_address, str)
+                if ip_address == "::1":
+                    continue
                 ipv6_address = ip_address
 
         if ipv6_address:


sambacc/commands/ctdb.py
```diff
@@ -271,17 +271,19 @@ def _lookup_hostname(hostname: str) -> str:
...
+                if ip_address.startswith("127."):
+                    continue
                 return ip_address
 
             if family == socket.AF_INET6 and ipv6_address is None:
+                ip_address = sockaddr[0]
+                assert isinstance(ip_address, str)
+                if ip_address == "::1":
+                    continue
                 ipv6_address = ip_address
```
anoopcs9, line 287: @avanthakkar Why are we not returning IPv6 address from here? What special case are we dealing with additional `ipv6_address`?

anoopcs9 (approved):
Fix itself lgtm.

See below for a comment on the existing address lookup logic.

sambacc/commands/ctdb.py
avanthakkar: > @avanthakkar Why are we not returning IPv6 address from here? What special case are we dealing with additional `ipv6_address`?

Yeah you're right, we probably don't need to explicitly check for types, probably this is enough?
```
..............
ip_address = sockaddr[0]
assert isinstance(ip_address, str)

if ip_address.startswith("127.") or ip_address == "::1":
    continue
            
return ip_address
..................
```

avanthakkar (approved):
LGTM

sambacc/commands/ctdb.py
anoopcs9: > ```
> ..............
> ip_address = sockaddr[0]
> assert isinstance(ip_address, str)
> 
> if ip_address.startswith("127.") or ip_address == "::1":
>     continue
>             
> return ip_address
> ..................
> ```

I've created #147 in the above manner but retaining the check for address info family type.

PR #147, opened 2025-02-14
commands: Simplify _lookup_hostname() in ctdb module

anoopcs9:
@avanthakkar Why are we not returning IPv6 address from here? What special case are we dealing with additional `ipv6_address`?

_Originally posted by @anoopcs9 in https://github.com/samba-in-kubernetes/sambacc/pull/146#discussion_r1953874786_

### DIFF
--- a/sambacc/commands/ctdb.py
+++ b/sambacc/commands/ctdb.py
@@ -269,25 +269,14 @@ def _lookup_hostname(hostname: str) -> str:
             family=socket.AF_UNSPEC,
             type=socket.SOCK_STREAM,
         )
-        ipv6_address = None
 
         for family, _, _, _, sockaddr in addrinfo:
-            if family == socket.AF_INET:
+            if family == socket.AF_INET or family == socket.AF_INET6:
                 ip_address = sockaddr[0]
                 assert isinstance(ip_address, str)
-                if ip_address.startswith("127."):
+                if ip_address.startswith("127.") or ip_address == "::1":
                     continue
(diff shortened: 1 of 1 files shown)


avanthakkar (approved):
LGTM

phlogistonjohn:
I had assumed the way the code was previously was that it was designed to prefer *any* ipv4 address over an ipv6 address. @avanthakkar is that not the case? - I see you approved already, and so I guess that is not important - but I would like to confirm this clearly.

anoopcs9:
> I had assumed the way the code was previously was that it was designed to prefer _any_ ipv4 address over an ipv6 address.

Ah..I see that this preference is clearly mentioned in the description for https://github.com/samba-in-kubernetes/sambacc/pull/137. I was searching for any indications in [commit message](https://github.com/samba-in-kubernetes/sambacc/commit/dc564c8c574157e6476ef6fdab05734790b41b8a) and there wasn't any.

So this change is no longer relevant.

avanthakkar:
> > I had assumed the way the code was previously was that it was designed to prefer _any_ ipv4 address over an ipv6 address.
> 
> Ah..I see that this preference is clearly mentioned in the description for #137. I was searching for any indications in [commit message](https://github.com/samba-in-kubernetes/sambacc/commit/dc564c8c574157e6476ef6fdab05734790b41b8a) and there wasn't any.
> 
> So this change is no longer relevant.

ah yes, my bad! I missed to take this point into consideration

PR #148, opened 2025-02-26
commands: handle /usr/etc/nsswitch.conf

jeffmahoney:
On some newer OS releases, standard config files are installed in /usr/etc with site changes to copies in /etc.

That results in the following startup failure:
Traceback (most recent call last):
  File "/usr/bin/samba-container", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "sambacc/commands/main.py", line 231, in main
    cfunc(CommandContext(cli))
  File "sambacc/commands/run.py", line 75, in run_container
    init_container(ctx)
  File "sambacc/commands/initialize.py", line 116, in init_container
    cmds[step_name].cmd_func(ctx)
  File "sambacc/commands/initialize.py", line 38, in _import_nsswitch
    nss.read()
  File "sambacc/textfile.py", line 28, in read
    with open(self.path) as f:
         ^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/etc/nsswitch.conf'

This changes the initialization so that we fall back to reading from /usr/etc/nsswitch.conf if reading from /etc/nsswitch.conf fails but still write out any changes to /etc/nsswitch.conf.

### DIFF
--- a/sambacc/textfile.py
+++ b/sambacc/textfile.py
@@ -28,11 +28,14 @@ class TextFileLoader:
         with open(self.path) as f:
             self.readfp(f)
 
-    def write(self) -> None:
-        tpath = self._tmp_path(self.path)
+    def write(self, alternate_path: typing.Optional[str] = None) -> None:
+        path = self.path
+        if alternate_path:
+            path = alternate_path
+        tpath = self._tmp_path(path)
         with open(tpath, "w") as f:
             self.writefp(f)
-        os.rename(tpath, self.path)
+        os.rename(tpath, path)
 
     def _tmp_path(self, path: str) -> str:
         # for later: make this smarter
--- a/sambacc/commands/initialize.py
+++ b/sambacc/commands/initialize.py
@@ -34,11 +34,19 @@ _logger = logging.getLogger(__name__)
 @setup_steps.command("nsswitch")
 def _import_nsswitch(ctx: Context) -> None:
     # should nsswitch validation/edit be conditional only on ads?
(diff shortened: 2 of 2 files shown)


sambacc/textfile.py
```diff
@@ -28,11 +28,13 @@ def read(self) -> None:
         with open(self.path) as f:
             self.readfp(f)
 
-    def write(self) -> None:
-        tpath = self._tmp_path(self.path)
+    def write(self, path: typing.Optional[str] = None) -> None:
```
phlogistonjohn, line 31: Most of this patch looks OK to me, but this part I find a bit confusing. The path is provided to the object at initialization time, and then again (optionally) here - that seems redundant to me.

jeffmahoney: On systems where this matters, /usr is typically read-only. Any changes to the config files in /usr/etc are written to /etc and prioritized over the ones in /usr/etc. This change allows us to write the contents of the read-in (and modified) file back out to a different location. When the path is unspecified, the existing behavior is preserved.

phlogistonjohn:
I'd like to see this move forward. Do you think you'll have time soon to fix the tests? (and maybe rename that argument :-) )  Thanks!

phlogistonjohn:
@jeffmahoney ping ^

jeffmahoney:
Ack. Changes made and tox run. Rebased on current HEAD to get rid of that merge commit.

phlogistonjohn (approved):
lgtm, thanks

spuiuk (approved):
ACK

PR #149, opened 2025-03-11
ctdb: skip mutex helper service registration

phlogistonjohn:
Depends on: https://gitlab.com/samba-team/samba/-/merge_requests/4002

When the ctdb ceph rados mutex helper tool supports the -R option to
skip registering the helper as a ceph service - do so.
This avoids confusion on the ceph side as the helper will act more like
a typical rados client than something on the level of an OSD or MDS.


avanthakkar (approved):
LGTM!

phlogistonjohn:
Yeah, that it is what I am looking into next.

PR #152, opened 2025-05-03
commands: invalidate cached config when node is not ctdb leader

phlogistonjohn:
When the `samba-container config-update --watch` command is run on a CTDB enabled server we avoid updating the local config unless the node is the ctdb leader. However, the update function was returning the current config which the watch loop assigned to the previous var, meaning that the current and previous config would be equal on future trips around the loop. Therefore once the node became the leader it would not update the config even if it differed from the config in samba. This is a problem on start up because the loop initially runs before ctdb is ready and the node leadership is not known. Change it so that when the node is not a leader it returns None and "unsets" the previous config. This could lead to extra unneeded updates but should avoid missing updates that are needed.

### DIFF
--- a/sambacc/commands/config.py
+++ b/sambacc/commands/config.py
@@ -79,12 +79,15 @@ def _read_config(ctx: Context) -> config.InstanceConfig:
     ).get(ctx.cli.identity)
 
 
+UpdateResult = typing.Tuple[typing.Optional[config.InstanceConfig], bool]
+
+
 def _update_config(
     current: config.InstanceConfig,
     previous: typing.Optional[config.InstanceConfig],
     ensure_paths: bool = True,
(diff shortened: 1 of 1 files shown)


anoopcs9:
What is the abnormality seen with the current code? Did we observe a cluster stuck in UNHEALTHY or some other undesired state?

phlogistonjohn:
The observed issue was that on the Ceph SMB system there's a brief race condition where if you create shares while the ceph orchestration is still rolling out the first set of containers it will apply the "initial config" to smb but the config watch sidecar will not do anything. There's a few more details in the commit message.

anoopcs9 (approved):
lgtm.

avanthakkar (approved):
LGTM!

PR #153, opened 2025-05-06
commands: a variety of small cli handling cleanups

phlogistonjohn:
In preparation for some future work, clean up how we construct the sambacc command line tools: samba-container and samba-dc-container. Make common stuff common, add docstrings and make things less special.


sambacc/commands/cli.py
```diff
@@ -45,13 +45,11 @@ class Parser(typing.Protocol):
 
     def set_defaults(self, **kwargs: typing.Any) -> None:
         """Set a default value for an argument parser."""
-        ...  # pragma: no cover
```
anoopcs9, line 48: Commit messages says only about ellipsis and not pragma. Should we retain it?

anoopcs9 (commented):
Can we use `commands.include()` or `commands.include_multiple()` for those imports with noqa comments in _initialize.py_?

sambacc/commands/cli.py
phlogistonjohn: The pramga is just there to tell coverage tool not to count the ellipis line when detecting uncovered lines. So that comment is only there because the `...` is there.

phlogistonjohn:
> Can we use `commands.include()` or `commands.include_multiple()` for those imports with noqa comments in _initialize.py_?

Not directly as I added the new method to handle the functions that are decorated by `commands` decorator. The imports in initialize.py are there to pull in the setup step decorators. We could add something similar in the future for that but I think I would not want to do that for this pr.

anoopcs9 (approved):
lgtm, thanks.

xhernandez (approved):

phlogistonjohn:
@mergifyio rebase

phlogistonjohn:
mergify seems to be to be ingoring Xavi's review because he didn't have write perms on this repo before today. I need one more review, or I can just force it by putting on the label... I may wait a little while... but not too long :-)

PR #155, opened 2025-05-27
mergify: Restrict the addition of "priority-review" label

anoopcs9:
Mergify rules doesn't check the status of pull requests before adding the "priority-review" label. In addition to the existing rules it only makes sense to add the label if the pull request is in open state.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -77,6 +77,8 @@ pull_request_rules:
     conditions:
       - "updated-at<15 days ago"
       - "-draft"
+      - "-closed"
+      - "-merged"
     actions:
       label:
         add:


obnoxxx (commented):
This sounds reasonable and I think it is a good idea.

looks good in general, but I doubt that `draft`is the same as `open`here.

.github/mergify.yml
```diff
@@ -77,6 +77,8 @@ pull_request_rules:
     conditions:
       - "updated-at<15 days ago"
       - "-draft"
+      - "-closed"
+      - "-merged"
```
obnoxxx, line 81: this leaves only draft.

anoopcs9: @obnoxxx Can you please explain if there is any real concern here?

phlogistonjohn:
@mgerify rebase

phlogistonjohn:
@Mergifyio rebase

PR #157, opened 2025-05-29
workflows: Dynamically detect fedora versions for test matrix

### LINKED ISSUE #145: automate running on latest fedora version in ci matrix
> I wonder if I should rename the fedora test jobs something like `fedora-previous` and `fedora-current` or something to avoid this (small bit of) churn? Any opinions?

Well, we can't get rid of changes unless we do something as follows:
```
latest=$(curl -s -L https://fedoraproject.org/releases.json | jq -r '[.[].version | select("^[0-9]+$")] | max')

previous=$(( $latest - 1 ))
```

And of course I prefer "latest" over "current". 

Would you like to go for it with the current PR?

_Originally posted by @anoopcs9 in https://github.com/samba-in-kubernetes/sambacc/pull/140#pullrequestreview-2450314857_

anoopcs9:
Instead of updating the versions during each fedora release we could automatically filter out the recent versions from publically available json data for fedora releases. We also take this opportunity to rename the corresponding test jobs to avoid updating the mergify configuration for every fedora release.

fixes #145

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -7,8 +7,8 @@ queue_rules:
   - name: default
     conditions:
       - check-success=check-commits
-      - check-success=test (fedora-41)
-      - check-success=test (fedora-40)
+      - check-success=test (fedora-latest)
+      - check-success=test (fedora-previous)
       - check-success=test (centos-stream9)
       - check-success=dpulls
     merge_method: rebase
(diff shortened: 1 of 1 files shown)


anoopcs9:
> Instead of labeling the jobs with the version number call them `fedora-latest` and `fedora-previous` (or something)? Then we do not need to keep updating that file.

Perfect, thanks.

> Or are you suggesting to dump mergify entirely? I would not be opposed but I want an equivalent workflow (via actions?) and I don't know if we want to take that on now.

Nah..

phlogistonjohn (approved):

xhernandez (approved):

anoopcs9:
@Mergifyio rebase

Issue #156, opened 2025-05-29
gRPC protobuf doesn't contain a field for Encryption in SessionInfo


anoopcs9:
I would say we add a field for Encryption.

_Originally posted by @anoopcs9 in https://github.com/samba-in-kubernetes/sambacc/pull/154#discussion_r2111441930_

PR #158, opened 2025-05-30
grpc: add session crypto information support

phlogistonjohn:
Fixes: #156 

After looking at the code and the implemenation in samba I decided it was easiest and best to mimic the structure seen in the JSON, adding a new SessionCrypto message type and using that for the encryption and signing fields of the smbstatus.


anoopcs9 (changes requested):
Apart from the typo in the commit subject, `grpc: add support for session crypto inforamation` could mention "backend" as it basically modifies the grpc backend.

anoopcs9 (approved):
lgtm, thanks.

Shwetha-Acharya (approved):
LGTM

Issue #141, opened 2024-12-02
samba-container check winbind fails when _GLOBAL_DEBUG is set


eivindlm:
When `_GLOBAL_DEBUG` is set to 10, `samba-container check winbind` attempts to exec `wbinfo -p --debuglevel=10`, which fails with "Invalid option". This will cause the livenessProbe to fail, effectively killing the container every 30 seconds.

phlogistonjohn:
Thank you very much for the report!

anoopcs9:
`wbinfo --ping --debuglevel=10` currently doesn't error out which seems to be a bug in the `wbinfo` tool itself whereas `wbinfo --debuglevel=10 --ping` is invalid. Luckily `samba-container check winbind` puts `--debuglevel` towards the end and thus it works(where it is not supposed to).

@eivindlm Can you please re-confirm whether `wbinfo --ping --debuglevel=10` displays "Invalid option" for you or not inside the container? If yes, can you paste the samba container image name and tag used?

anoopcs9:
> My wbinfo -p still exits with nonzero return if additional arguments (such as --debuglevel=10) are provided.
> 
> ```
> [me@my-host ~]$ kubectl -n samba-shares exec -it my-pod-0 -c wb -- /bin/bash
> [root@my-pod-0 /]# wbinfo --version
> Version 4.20.5
> [root@my-pod-0 /]# wbinfo --ping
> Ping to winbindd succeeded
> [root@my-pod-0 /]# echo $?
> 0
> [root@my-pod-0 /]# wbinfo --ping --debuglevel=10
> Ping to winbindd succeeded
> Invalid option
> Usage: [OPTION...]
> [ ..... several lines of output omitted .....]
> [root@my-pod-0 /]# echo $?
> 1
> ```

Thanks, it is interesting to see that the message indicating a successful ping is also displayed at the top before "Invalid option".

Nevertheless, I think it is wise to not execute the underlying **wbinfo** command with `--debuglevel` optional argument. I'll create a fix.

PR #159, opened 2025-06-04
samba_cmds: Do not add --debuglevel for `wbinfo` and `ctdb` commands

anoopcs9:
**wbinfo** and **ctdb** commands do not accept `--debuglevel` optional argument. Make sure that we execute those commands without `--debuglevel` in the presence of `--samba-debug-level` optional argument to **samba-container** command or its equivalent with `SAMBA_DEBUG_LEVEL` environment variable.

fixes https://github.com/samba-in-kubernetes/sambacc/issues/141

### DIFF
--- a/sambacc/ctdb.py
+++ b/sambacc/ctdb.py
@@ -764,12 +764,12 @@ def archive_tdb(iconfig: config.InstanceConfig, dest_dir: str) -> None:
                 os.rename(tdb_path, dest_path)
 
 
-def check_nodestatus(cmd: samba_cmds.SambaCommand = samba_cmds.ctdb) -> None:
+def check_nodestatus(cmd: samba_cmds.CommandArgs = samba_cmds.ctdb) -> None:
     cmd_ctdb_check = cmd["nodestatus"]
     samba_cmds.execute(cmd_ctdb_check)
 
 
-def _read_command_pnn(cmd: samba_cmds.SambaCommand) -> typing.Optional[int]:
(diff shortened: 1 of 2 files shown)


sambacc/commands/check.py
```diff
@@ -34,7 +36,7 @@ def check(ctx: Context) -> None:
     """Check that a given subsystem is functioning."""
     if ctx.cli.target == "winbind":
         cmd = samba_cmds.wbinfo["--ping"]
-        samba_cmds.execute(cmd)
+        os.execvp(cmd.name, cmd.argv())
```
phlogistonjohn, line 39: There's already a execvp wrapper in the samba_cmds.py file: https://github.com/samba-in-kubernetes/sambacc/blob/b6e11822764dc7e31472d0f93910ab42417ac4f8/sambacc/samba_cmds.py#L212
Is there a reason you didn't use it?

phlogistonjohn: [Game show host voice] So close, but I'm afraid the answer is CommandArgs. The judges would also have accepted a new protocol, but typing.Any is too broad. [/joke] :-)

CommandArgs is the right choice because it is the base class of SambaCommand, thus all SambaCommands are CommandArgs. Annotating it with typing.Any is too general and would have the type checker alloing things like `execute(7)` which would fail at runtime because an int doesn't have a .name or .argv method.

anoopcs9: Ah..I'm totally exposed now, just kidding.

Updated.

avanthakkar (approved):
lgtm

PR #161, opened 2025-06-11
ctdb: add option to change ctdb port

phlogistonjohn:
Add a mechanism for changing the CTDB port.


sambacc/ctdb.py
```diff
@@ -181,6 +182,28 @@ def read_ctdb_nodes(path: str = CTDB_NODES) -> list[str]:
...
+
+def ensure_ctdb_port_in_services(port: int, path: str) -> None:
+    try:
+        with open(path, "r") as fh:
+            lines = [line.strip() for line in fh]
+    except FileNotFoundError:
+        lines = []
+    cleaned = [line for line in lines if not _svc_match("ctdb", line)]
+    cleaned.append(f"ctdb  {port}/tcp   # custom ctdb port")
+    cleaned.append(f"ctdb  {port}/udp   # custom ctdb port")
```
spuiuk, line 200: Do we need an entry for udp here? According to https://wiki.samba.org/index.php/Advanced_CTDB_configuration#CTDB_port, it only opens tcp ports.

spuiuk (approved):
Minor change requested. 

But otherwise ACK.

sambacc/ctdb.py
phlogistonjohn: The default file has both entries. On my system (fedora 41):
```
[jmulliga@nacho ~]$ grep ctdb /etc/services 
ctdb            4379/tcp                # CTDB
ctdb            4379/udp                # CTDB
```
I wanted it to match.

PR #162, opened 2025-06-23
add files to include in smb.conf stub for ctdb 

phlogistonjohn:
When using ctdb most of the configuration should be managed via the
registry config via ctdb. However, there are some parameters you may
need to be per-instance. As a quick and simple way of getting custom
per host settings with ctdb allow for including additional files from
the generated stub smb.conf file.

Plus two minor cleanups noticed while hacking on this feature.


anoopcs9 (approved):
lgtm, thanks.

avanthakkar (approved):
LGTM!

PR #178, opened 2025-10-28
sambacc: use tempfile as arg to net conf import

spuiuk:
net conf import is unable to use the STDIN to import the configuration when the generated smb.conf is large. Use a tempfile instead as an argument to net conf import.

Resolves: rhbz#2400102

### DIFF
--- a/sambacc/netcmd_loader.py
+++ b/sambacc/netcmd_loader.py
@@ -18,6 +18,7 @@
 
 import subprocess
 import typing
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
https://github.com/samba-in-kubernetes/sambacc/pull/179 for a fix for the ci failure

anoopcs9:
@Mergifyio rebase

PR #179, opened 2025-10-28
commands: work around a new flake8 failure

phlogistonjohn:
Suddenly, on the new f43 image ci run flake8 fails. Looking at the error and the code I think this failure is incorrect but I don't want to block the CI and wait for flake8 to be changed. The workaround is low cost and not much harder to understand than the original code.


xhernandez (commented):
lgtm

PR #180, opened 2025-11-05
sighup pids in pidfiles when config changes

phlogistonjohn:
Fix issues with the WIP effort to add fscrypt/keybridge support to SMB on Ceph:

The issue is that the keybridge configuration on ceph is changed but the keybridge sidecar just keeps using the configuration data in place when the process started. The patches in this PR:
* Add support for writing process pids to a pidfile
* Add ability for config watch to pick up the pidsfiles in a given dir and SIGHUP them all
* Ensures that keybridge configuration is considered when comparing old and new configurations in config watch
* Fixes instance configuration being cached on the CommandContext between restarts of the keybridge sidecar

This has been manually tested in tandem with WIP changes to cephadm to enable the new pidfile related options for keybridge and config watch.

### DIFF
--- a/sambacc/commands/satellite/keybridge.py
+++ b/sambacc/commands/satellite/keybridge.py
@@ -26,6 +26,7 @@ import typing
 import sambacc.config
 
 from ..cli import Context, Fail, commands
+from ..common import CommandContext
 
 
 if typing.TYPE_CHECKING:
@@ -129,6 +130,7 @@ class Restart(Exception):
 @commands.command(name="keybridge", arg_func=_serve_args)
 def serve_keybridge(ctx: Context) -> None:
     """Start a keybridge varlink RPC server."""
+    assert isinstance(ctx, CommandContext)  # for copy method
 
     def _handler(*args: typing.Any) -> None:
         raise Restart()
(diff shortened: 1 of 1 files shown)


sambacc/commands/config.py
```diff
@@ -141,14 +149,81 @@ def _call_if_leader(
...
+        _pids = [
+            _readpid(p)
+            for p in pathlib.Path(pids_dir).iterdir()
+            if p.name.endswith(".pid")
+        ]
+    except FileNotFoundError:
+        _logger.warning("pids dir not found: %r; skipping pids", pids_dir)
+        return current, changed
+
+    pids = [pid for pid in _pids if pid > 0]  # filter out errors
```
synarete: Maybe change to `pid > 1` (pid number 1 is init in Linux)

synarete (approved):
lgtm

sambacc/commands/config.py
phlogistonjohn: In containers pid one could be a sambacc component or smbd. It probably won't be in ceph (IIRC it should be a podman init component) but I would rather not exclude a valid but special pid because I do want the feature reusable on (theoretical) non-ceph deployments.

PR #182, opened 2026-01-29
remote-control: extend grpc API with config dump endpoints 

phlogistonjohn:
To improve and streamline certain workflows add new API endpoints for getting configuration dumps for diagnostic purposes.

The new API endpoints:
* ConfigDump: for listing underlying configurations
* ConfigSummary: for generating a digest of configurations
* ConfigSharesList: for generating a list of shares

These apis can be applied to Samba (via net conf), sambacc, and ctdb.
The ctdb configuration will not support ConfigSharesList.

The ConfigDump API should be fairly obvious in it's purpose, get the configuration so it can be examined (or logged and examined later). Because of the potential length of a configuration the API streams objects containing a line number and the content of a line.

The ConfigSummary is used to create a "short" digest of the configuration that can serve as a fingerprint to detect is configuration has changed. The ConfigSummary API takes an (optional) hash algorithm. Currently only sha256 is supported.

To avoid repeat API calls, the ConfigDump API also takes an optional hash algorithm. If specified, the final item in the return stream will contain a digest field. One could use this apis to get the config dump + hash digest and then use the ConfigSummary call to query if the config has changed.

The ConfigSharesList exists to get a quick summary of the shares without using the smb protocol itself and it always lists all shares regardless of "browsability" in the smb conf.

### DIFF
--- a/tests/test_grpc_server.py
+++ b/tests/test_grpc_server.py
@@ -126,6 +126,31 @@ class MockBackend:
     def kill_client(self, ip_address: str) -> None:
         self._counter["kill_client"] += 1
 
+    def config_dump(self, source, hash_alg):
+        self._counter["config_dump"] += 1
+        if self._kaboom:
+            raise self._kaboom
+        yield backend.DumpItem(line_number=0, content="foo\n")
+        yield backend.DumpItem(line_number=0, content="bar\n")
+        yield backend.DumpItem(line_number=0, content="baz\n")
+        yield backend.DumpItem(line_number=0, content="bingo\n")
+        if hash_alg:
+            yield backend.DumpItem(
+                line_number=-1, hash_type="sha256", content="xxxxxxxx"
+            )
+
+    def config_dump_digest(self, source, hash_alg):
+        self._counter["config_dump_digest"] += 1
+        return backend.DumpItem(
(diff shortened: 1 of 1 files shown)


sambacc/grpc/generated/control_pb2.pyi
```diff
@@ -8,16 +8,57 @@ import builtins
...
 
 DESCRIPTOR: google.protobuf.descriptor.FileDescriptor
 
+class _ConfigFor:
+    ValueType = typing.NewType("ValueType", builtins.int)
+    V: typing_extensions.TypeAlias = ValueType
+
+class _ConfigForEnumTypeWrapper(google.protobuf.internal.enum_type_wrapper._EnumTypeWrapper[_ConfigFor.ValueType], builtins.type):  # noqa: F821
+    DESCRIPTOR: google.protobuf.descriptor.EnumDescriptor
+    CONFIG_FOR_MISSING: _ConfigFor.ValueType  # 0
```
avanthakkar: Maybe `CONFIG_FOR_UNSET` similar to `HASH_ALG_UNSET` ?

avanthakkar (approved):
LGTM

PR #186, opened 2026-02-05
remote-control: extend grpc API with debug level endpoints

phlogistonjohn:
To make live debugging easier, add new endpoints that support
a (mostly) uniform interface for setting and getting the debug
logging level of samba processes.

The new API endpoints:
* SetDebugLevel: for modifying the current debug level of a process
* GetDebugLevel: for querying the current debug level of a process

These APIs can be used against smbd, windbindd, and ctdbd by setting
the appropriate field in the function request object. Both functions
return an info object containing the process and the new/current level.

The level is supplied as a string and the value depends on the process.
Numbers for smbd and winbindd and words (like "ERROR" or "WARNING") for
ctdb.

### DIFF
--- a/tests/test_grpc_server.py
+++ b/tests/test_grpc_server.py
@@ -151,6 +151,13 @@ class MockBackend:
         yield backend.ShareEntry(name="bob")
         yield backend.ShareEntry(name="zongo")
 
+    def set_debug_level(self, server, debug_level):
+        self._counter["set_debug_level"] += 1
+
+    def get_debug_level(self, server):
+        self._counter["get_debug_level"] += 1
+        return "ERROR" if server is backend.ServerType.CTDB else "10"
+
 
 @pytest.fixture()
 def mock_grpc_server(tmp_path):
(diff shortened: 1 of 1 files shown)


sambacc/grpc/generated/control_pb2_grpc.py
```diff
@@ -106,6 +117,22 @@ def ConfigSharesList(self, request, context):
...
+        """---- live debug level control ----
+
+        SetDebugLevel attempts to set the debug level for an smb process.
+        """
+        context.set_code(grpc.StatusCode.UNIMPLEMENTED)
+        context.set_details('Method not implemented!')
+        raise NotImplementedError('Method not implemented!')
+
+    def GetDebugLevel(self, request, context):
+        """SetDebugLevel attempts to get the debug level for an smb process.
```
avanthakkar:
```suggestion
        """GetDebugLevel attempts to get the debug level for an smb process.
```

sambacc/grpc/server.py
```diff
@@ -257,6 +275,34 @@ def ConfigSharesList(
...
+            self._backend.set_debug_level(server, debug_level)
+            # round trip for clarity
+            info = pb.DebugLevelInfo(
+                process=request.process,
+                debug_level=debug_level,
+            )
+        return info
+
+    def GetDebugLevel(
+        self, request: pb.SetDebugLevelRequest, context: grpc.ServicerContext
```
avanthakkar:
```suggestion
        self, request: pb.GetDebugLevelRequest, context: grpc.ServicerContext
```

PR #187, opened 2026-02-13
ceph: create a new ceph dir and move rados_opener to ceph.rados

phlogistonjohn:
Create a new namespace for ceph integrations in sambacc. Move the existing rados_opener to ceph.rados, as the module is doing far more than just supporting an opener anyway. This is in preparation for doing even more integration with ceph and rados for various parts of sambacc.


synarete (approved):
lgtm

PR #209, opened 2026-07-02
sambacc: always start watch loop with previous=None

### LINKED ISSUE #208: update-config --watch skips initial registry sync
When samba-container `update-config --watch` starts, it uses the already loaded `ctx.instance_config` as its baseline previous value. 
Since `_get_config()` in the watch loop reads the same URIs and produces the same result, the first iteration sees no change and skips writing to the samba registry. This means the registry is never updated with values from `config:merge` stubs in secondary `SAMBACC_CONFIG URIs`.

avanthakkar:
`update-config --watch` passed `ctx.instance_config` as the initial previous into the watch loop. Because `ctx.instance_config` is loaded from the same full `SAMBACC_CONFIG` URI list (including config:merge stubs) as `_get_config()`, the first iteration always evaluated no change and the registry was never written on startup.

When a secondary `rados:mon-config-key` URI carrying a `config:merge` stub was added to `SAMBACC_CONFIG` (e.g. Ceph RGW credentials), the merged values never made it into the registry. `net conf list` showed empty, while 
`samba-container print-config` showed the correct merged output. 

This PR intends to fix this issue by passing `previous = None` for watch path as well which will now unconditionally trigger a registry write on the first iteration. 

Fixes: https://github.com/samba-in-kubernetes/sambacc/issues/208

### DIFF
--- a/sambacc/commands/config.py
+++ b/sambacc/commands/config.py
@@ -232,9 +232,12 @@ def update_config(ctx: Context) -> None:
     if ctx.cli.watch:
         _logger.info("will watch configuration source")
         waiter = best_waiter(ctx.cli.config)
+        # Pass None as initial_value so the first iteration always writes the
+        # current config to the samba registry, consistent with the non-watch
+        # code path which also passes None unconditionally.
         watch(
             waiter,
-            ctx.instance_config,
+            None,
             _get_config,
             trigger,
         )
--- a/tests/test_commands_config.py
+++ b/tests/test_commands_config.py
@@ -227,6 +227,10 @@ def test_update_config_ctdb_notleader(tmp_path, monkeypatch):
 
 
 def test_update_config_watch_waiter_expires(tmp_path, monkeypatch):
+    # Verifies that watch mode performs an initial registry sync on the first
+    # iteration even when the config file hasn't changed since the process
+    # started.  Subsequent iterations with no config change must not trigger
+    # additional writes (only the initial sync happens).
     cfg_path = str(tmp_path / "config")
     fake = tmp_path / "fake.sh"
     chkpath = tmp_path / ".executed"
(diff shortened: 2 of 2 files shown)


phlogistonjohn:
I dunno. I feel a bit squeamish about changing something that fundamental. Should we gate this behavior change behind a cli option?

phlogistonjohn:
--reload-on-start ?

avanthakkar:
> I dunno. I feel a bit squeamish about changing something that fundamental. Should we gate this behavior change behind a cli option?

I honestly felt sketchy about changing watcher API itself but so far I couldn't found workaround to it because I think the core issue is that the samba registry is empty on every container restart, including after a RECONFIG, which doesn't re-run `import-config`. 
The non-watch path already passes `None` unconditionally, so watch mode was inconsistent with it.

Simplest fix that avoids touching `watch()`: just pass `None` at the call site in `update_config` instead of `ctx.instance_config`. Wdyt?

phlogistonjohn:
Yeah, let's do that.

phlogistonjohn:
(and a comment :-))

avanthakkar:
> Yeah, let's do that.

Done!

phlogistonjohn (approved):

phlogistonjohn:
I approved before fully reading the comment:
> The registry is empty after  every container restart

I dont think this is always true... so can we just be more vague and just leave out this bit?

phlogistonjohn (changes requested):
.

avanthakkar:
> I approved before fully reading the comment:
> 
> > The registry is empty after  every container restart
> 
> I dont think this is always true... so can we just be more vague and just leave out this bit?

Ah yes, correct! Removed that part

phlogistonjohn (approved):

anoopcs9 (approved):
lgtm, thanks.

PR #211, opened 2026-07-14
ceph/rados: close cluster-meta connection on context exit

### LINKED ISSUE #210: ctdb-monitor-nodes leaks a librados connection every poll iteration -> unbounded thread/anon-memory growth (host OOM)
## Environment

- sambacc `0.9.dev58+g855773d73` (commit `855773d73`, 2026-04-06). The relevant paths — `sambacc/ceph/rados.py`, `sambacc/ctdb.py`, `sambacc/commands/ctdb.py`, `simple_waiter.py`, `inotify_waiter.py` — are **byte-identical to current `master` (`c989a9476`)**, so this reproduces on latest.
- Container image: `quay.io/samba.org/samba-server:ceph20-centos-amd64`. **Reproduced on manifest digest `sha256:04d481b8efd5bd7ff45f5b07376376dde1b09c2eb299e1a802a4e0c166e36f47` (current 2026-07-12)**, which ships `python3-sambacc-0.10~4.gc989a9476` — i.e. a build from current `master` head (`c989a9476`). So this is present on the latest image, not just an older pinned build.
- Samba 4.23.5, ctdb 4.23.5.
- Deployed by Ceph **v20.2.2 (Tentacle)** cephadm SMB mgr module; clustered/CTDB (`instance_features: ["ctdb"]`); 2 nodes, `auth_mode: user`, `clustering: default`.

## What happens

The long-lived `samba-container … ctdb-monitor-nodes` process accumulates threads and anonymous memory without bound. Captured on two nodes:

| | node A (heavier) | node B |
|---|---|---|
| Threads (nlwp) | 45,745 | 36,325 |
| RSS | ~23.8 GiB | ~17.2 GiB |
| VmSize | ~392 GiB | ~317 GiB |
| Uptime at capture | ~63 h | ~50 h |

- **Growth is exactly +12 threads per ~65 s poll**, monotonic (sampled live: 45757 → 45769 → 45781). 45,744 non-main threads ÷ 12 = **3,812 accumulated groups**.
- Each group is a complete **librados/messenger thread-team**: 3× `msgr-worker`, `ms_dispatch`, `ms_local`, `log`, `ceph_timer`, `service`, 2× `safe_timer`, 2× `io_context_pool`. One new RADOS client per poll, never shut down.
- **100 % of threads are parked** (`State: S`), kernel stacks at `futex_wait` / `ep_poll` — never joined, doing no work.
- Memory is the leaked stacks: VmData/thread ≈ 8.57 MB (glibc default 8 MB pthread stack); main-thread VmStk only 764 kB, so VmSize is essentially all thread-stack reservation, ~0.5 MB RssAnon touched per stack.
- The owning `ctdbNodes` container cgroup is **uncapped** (`memory.max = max`), so growth (~8–9 GiB/day) proceeds to a **global** host OOM. We observed a prior generation OOM-kill the host (kernel log: `msgr-worker-2 invoked oom-killer … global_oom … Killed process … samba-container … total-vm:596067776kB`), after which cephadm restarted the sidecar and the leak resumed — a repeating cycle. **Ceph reports `HEALTH_OK` throughout**, so this is silent until the node dies.

## How the daemon is invoked

cephadm renders the ctdbNodes sidecar with two flags hardcoded unconditionally in `CTDBNodeMonitorContainer.args()` (`cephadmlib/daemons/smb.py`, Ceph v20.2.2):

```
samba-container --debug ctdb-monitor-nodes \
  --hostname=smb.smb-media.0.1.ceph-2.panvjs \
  --take-node-number-from-env --reload=all
```

`--debug` and `--reload=all` are static list literals (no config gate); operators cannot change this. Runtime argv confirmed via `docker inspect`.

## Minimal reproduction (isolates the bug — no CTDB/samba harness needed)

Run inside the current `ceph20-centos-amd64` image (digest above) on a host with a `ceph.conf` + keyring, against a **dummy** object so nothing is read/written and no lock is taken:

```python
import os
from sambacc.ceph.rados import ClusterMetaRADOSObject

cmeta = ClusterMetaRADOSObject.create_from_uri(
    "rados://.smb/leakprobe-test/dummy-does-not-exist")  # scratch ns, nonexistent obj

for i in range(25):
    with cmeta.open(locked=False):   # locked=False → no CTDB lock; must_exist defaults False → no stat/read
        pass
    print(i + 1, len(os.listdir("/proc/self/task")))  # live thread count
```

Thread count climbs **exactly +12 every iteration** (1 → 301 over 25 opens):

```
1 13     2 25     3 37   ...   24 289   25 301
```

The end-of-run `/proc/self/task/*/comm` histogram shows each of the 25 leaked connections contributed a full librados/messenger thread-team — 3× `msgr-worker`, `ms_dispatch`, `ms_local`, `ceph_timer`, `service`, `log`, 2× `io_context_pool`, 2× `safe_timer` = 12 — none reclaimed. Each `open()` connects a new `rados.Rados()`; `ClusterMetaRADOSHandle.__exit__` never calls `close()`, so `shutdown()` is never invoked.

## Suspected root cause (source read at commit `855773d73`, == master; re-verified in the current image's on-disk `rados.py`)

In `sambacc/ceph/rados.py`, `ClusterMetaRADOSObject.open()` (~line 367) builds a fresh `RADOSObjectRef`, whose `_open()` (~lines 120–125) creates and **connects a new `rados.Rados()` handle every call** (spawning the native librados threads). The teardown path `RADOSObjectRef.close()` (~lines 185–190) is the only thing that calls `self._ioctx.close()` + `self._conn.shutdown()` — but the context manager `ClusterMetaRADOSHandle.__exit__` (~lines 354–359) **only releases the CTDB lock and returns; it never calls `close()`**. The connected handle is dropped and left to CPython GC / `__del__`, which for librados does not reliably run `shutdown()`.

`ctdb.monitor_cluster_meta_changes()` opens this handle via `with cmeta.open(locked=True)` on **every** poll iteration (including the unchanged / `continue` path), so each poll leaks one fully-connected librados handle (~10+ native threads + multi-MB buffers each). The author's own `# TODO: connection caching` at rados.py:121 flags these connections weren't meant to be created-and-dropped per call.

Note `--reload=all` is *not* itself the leaker: with it, `leader_locator` is `None` and reload is a synchronous `subprocess.check_call(ctdb reloadnodes)` that only fires on an actual node change — the steady leak is the per-iteration `open()`.

## Confirmation it is not already fixed

No existing sambacc / samba-container issue or PR describes a leak in this path (all issues/PRs reviewed; PR #127 introduced the command, PR #131 is a log-only fix; #208 is an unrelated `update-config --watch` fix). **The reproduction above was run on a build from current `master` head (`c989a9476`, sambacc `0.10~4.gc989a9476`), so this is present on latest — pulling a newer image does not fix it.**

## Workaround

No supported mgr-module knob caps the container or bounds the growth. Manual `ceph orch restart smb.smb-media` clears the accumulated threads/memory, but the leak restarts immediately and re-approaches OOM within days.

## Proposed fix

Minimal fix: have `ClusterMetaRADOSHandle.__exit__` close the underlying `RADOSObjectRef` (so `_ioctx.close()` + `_conn.shutdown()` run) in a `finally`, still releasing the lock — plus make `RADOSObjectRef.close()` idempotent. I have this patch and a regression test that fails before / passes after (asserts the connection is shut down once per `open()`-context exit, so connections don't accumulate across iterations), passing existing tests + `flake8`/`black`. Happy to open a PR if that's welcome. The larger alternative is the `# TODO: connection caching` — reuse one long-lived Rados handle across poll iterations.

dwburks:
## Summary

`ClusterMetaRADOSHandle.__exit__` never releases the RADOS connection it
opened. Every `ClusterMetaRADOSObject.open()` builds a brand-new, fully
`connect()`-ed `RADOSObjectRef` (one `rados.Rados()` client, which spawns
roughly a dozen librados threads), but the handle's context-manager exit only
drops the CTDB lock and returns. `RADOSObjectRef.close()` -- which calls
`ioctx.close()` and `conn.shutdown()` -- is therefore never invoked.

Because `ctdb.monitor_cluster_meta_changes()` opens this handle on every poll
(`with cmeta.open(locked=True) as cmo:`, ~once per minute, indefinitely), the
process leaks one fully-connected librados client per iteration. Threads and
memory grow without bound until the container is OOM-killed.

This PR makes the handle close its underlying object ref on exit, and makes
`RADOSObjectRef.close()` idempotent so it cannot double-`shutdown()`.

## Root cause

In `sambacc/ceph/rados.py`:

`ClusterMetaRADOSObject.open()` -> `_RADOSHandler.get_object()` ->
`RADOSObjectRef._from_uri(interface=self._interface, ...)`. Since `_interface`
is a `_RADOSInterface` (not a shared `RADOSConnection`), the ref takes the
`_open()` path:

```python
def _open(self, interface: _RADOSInterface) -> None:
    # TODO: connection caching
    self._api = interface.api
    self._conn = interface.Rados()
    self._conn.connect()
    self._connected = True
```

so each `open()` = one new connected client. The matching teardown exists on
the ref:

```python
def close(self) -> None:
    if not self._closed:
        self._ioctx.close()
        self._closed = True
    if self._connected:
        self._conn.shutdown()
```

but nothing calls it. The handle's exit is a no-op beyond unlocking:

```python
def __exit__(self, exc_type, exc_val, exc_tb) -> None:
    if self._locked:
        self._rados_obj._unlock(self._lock_name, self._cookie)
    return
```

No caller keeps the handle/ref alive past its `with` block. In `ctdb.py`
every consumer -- `monitor_cluster_meta_changes`, `add_node_to_cluster_meta`,
`refresh_node_in_cluster_meta`, `pnn_in_cluster_meta`, `_node_check`,
`_node_update`, `cluster_meta_to_nodes` -- uses the value returned by
`load()`/the effect of `dump()` and never touches the handle after the block.
Closing on exit is therefore safe: nothing relies on the connection persisting.

## The fix

`sambacc/ceph/rados.py`:

1. `ClusterMetaRADOSHandle.__exit__` now unlocks (when locked) inside a
   `try`, and closes the underlying ref in a `finally`. The `finally`
   guarantees the connection is freed even if `_unlock()` raises or the
   with-body raised (e.g. a failing `load()`):

   ```python
   def __exit__(self, exc_type, exc_val, exc_tb) -> None:
       try:
           if self._locked:
               self._rados_obj._unlock(self._lock_name, self._cookie)
       finally:
           self._rados_obj.close()
       return
   ```

2. `RADOSObjectRef.close()` now clears `self._connected` after shutting the
   connection down, so a second `close()` cannot call `shutdown()` twice:

   ```python
   if self._connected:
       self._conn.shutdown()
       self._connected = False
   ```

This is intentionally minimal and surgical -- it does not change how or when
connections are created, only that they are torn down when the handle's
context ends.

### Alternative / complementary direction

The connection churn itself (a fresh connect per poll) is the deeper cost; the
`# TODO: connection caching` note on `RADOSObjectRef._open()` points at the
better long-term fix: have `monitor_cluster_meta_changes()` (and friends) hold
one long-lived `RADOSConnection` and reuse it across iterations via the
existing `RADOSConnection.get_object()` / `_connected_to()` path, so each poll
reuses the socket instead of reconnecting. That is a larger, behavior-changing
refactor. This PR deliberately ships the small correctness fix -- stop leaking
-- and leaves connection caching as a follow-up.

## Reproduction

Observed on the master-head container image
`quay.io/samba.org/samba-server` digest `sha256:04d481b8...`, package version
`sambacc 0.10~4.gc989a9476` (repo commit `c989a9476`). Driving the exact
`monitor_cluster_meta_changes` open path in a loop:

```python
for _ in range(25):
    obj = ClusterMetaRADOSObject.create_from_uri("rados://<pool>/<ns>/<key>")
    with obj.open(locked=False) as cmo:
        cmo.load()
```

Thread count climbs monotonically by ~12 per iteration (one librados client's
worth), never dropping -- confirming a per-open connection leak. On a
long-running `ctdb-monitor-nodes` process this ends in host OOM. After the fix
the count is flat across iterations.

## Tests

`tests/test_rados_leak.py` (mirrors the mocking style of
`tests/test_rados_opener.py`; no real cluster required -- a `MagicMock`
stands in for the `rados` module and counts `connect`/`shutdown`/`close`):

- `test_cluster_meta_open_closes_connection` -- one `open()` cycle connects
  once and, on exit, shuts down + closes the ioctx exactly once.
- `test_cluster_meta_open_does_not_leak_across_iterations` -- the monitor-loop
  pattern over 25 iterations keeps `connect` and `shutdown` counts balanced
  (25 == 25). Fails on the current code (25 connects, 0 shutdowns).
- `test_cluster_meta_open_closes_when_body_raises` -- an exception in the
  with-body still unlocks and shuts the connection down.
- `test_rados_object_ref_close_is_idempotent` -- three `close()` calls result
  in exactly one `shutdown()`.

All four fail against `c989a9476` unpatched and pass with the fix applied.
The existing `tests/test_rados_opener.py` and `tests/test_ctdb.py` continue to
pass (31 passed total). `flake8` and `black --line-length 79` are clean on
both changed files.

## Verified on a live production node
In addition to the unit test, the fix was verified on a live cephadm SMB (CTDB) node (Ceph v20.2.2, image `quay.io/samba.org/samba-server:ceph20-centos-amd64`). With the patched `rados.py` bind-mounted over the installed module, an observer calling `ClusterMetaRADOSObject.open(locked=False)` every 5s held a **flat 1 thread across 74 iterations**, while an unpatched observer on the same node climbed **1 → 889 (+12 threads/iteration)** and the node's real `ctdb-monitor-nodes` daemon leaked at the **same +12/~68s** over the same window. The patch eliminates the leak.

Fixes #210

### DIFF
--- a/sambacc/ceph/rados.py
+++ b/sambacc/ceph/rados.py
@@ -188,6 +188,7 @@ class RADOSObjectRef(typing.IO):
             self._closed = True
         if self._connected:
             self._conn.shutdown()
+            self._connected = False
 
     @property
     def closed(self) -> bool:
@@ -354,8 +355,18 @@ class ClusterMetaRADOSHandle:
     def __exit__(
         self, exc_type: ExcType, exc_val: ExcValue, exc_tb: ExcTraceback
     ) -> None:
-        if self._locked:
-            self._rados_obj._unlock(self._lock_name, self._cookie)
+        try:
+            if self._locked:
+                self._rados_obj._unlock(self._lock_name, self._cookie)
+        finally:
+            # Always release the underlying RADOS connection. open() builds a
+            # fresh, fully-connected RADOSObjectRef for every context and no
+            # caller retains it past the `with` block, so failing to close
+            # here leaks one librados client (and its ~dozen threads) per
+            # invocation -- e.g. once per poll in
+            # ctdb.monitor_cluster_meta_changes(). Runs even if _unlock (or
+            # the with-body) raised; close() is idempotent.
+            self._rados_obj.close()
         return
 
 
--- a/tests/test_rados_leak.py
+++ b/tests/test_rados_leak.py
@@ -1,0 +1,102 @@
+#
+# sambacc: a samba container configuration tool
+# Copyright (C) 2026  John Mulligan
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>
+#
+
+import unittest.mock
+
+import pytest
+
+import sambacc.ceph.rados
+
+# CAUTION: these tests mock the ceph rados API. The fake Rados connection
+# counts connect/shutdown so we can assert that connections are released
+# rather than accumulated.
+
+
+@pytest.fixture(scope="function")
+def fake_cluster_meta(monkeypatch):
+    mock = unittest.mock.MagicMock()
+    # empty object body -> load() returns {}
+    mock.Rados.return_value.open_ioctx.return_value.read.return_value = b""
+    monkeypatch.setitem(sambacc.ceph.rados._module, "rados", mock)
+
+    interface = sambacc.ceph.rados._RADOSInterface()
+    interface.api = mock
+    interface.client_name = ""
+    interface.full_name = False
+    monkeypatch.setattr(
+        sambacc.ceph.rados._RADOSHandler, "_interface", interface
+    )
+
+    cmo = sambacc.ceph.rados.ClusterMetaRADOSObject.create_from_uri(
+        "rados://foo/bar/baz"
+    )
+    conn = mock.Rados.return_value
+    ioctx = conn.open_ioctx.return_value
+    return cmo, conn, ioctx
+
+
+def test_cluster_meta_open_closes_connection(fake_cluster_meta):
+    cmo, conn, ioctx = fake_cluster_meta
+
+    with cmo.open() as handle:
+        assert handle.load() == {}
+
+    assert conn.connect.call_count == 1
+    assert conn.shutdown.call_count == 1
+    assert ioctx.close.call_count == 1
+
+
+def test_cluster_meta_open_does_not_leak_across_iterations(fake_cluster_meta):
+    # ctdb.monitor_cluster_meta_changes() reopens once per poll; every open
+    # connects a fresh client, so connects and shutdowns must stay balanced.
+    cmo, conn, ioctx = fake_cluster_meta
+
+    iterations = 25
+    for _ in range(iterations):
+        with cmo.open(locked=True) as handle:
+            handle.load()
+
+    assert conn.connect.call_count == iterations
+    assert conn.shutdown.call_count == iterations
+    assert ioctx.close.call_count == iterations
+
+
+def test_cluster_meta_open_closes_when_body_raises(fake_cluster_meta):
+    cmo, conn, ioctx = fake_cluster_meta
+
+    with pytest.raises(RuntimeError):
+        with cmo.open(locked=True):
+            raise RuntimeError("boom")
+
+    assert ioctx.unlock.called
+    assert conn.shutdown.call_count == 1
+    assert ioctx.close.call_count == 1
+
+
+def test_rados_object_ref_close_is_idempotent():
+    mock = unittest.mock.MagicMock()
+    rr = sambacc.ceph.rados.RADOSObjectRef(mock, "foo", "bar", "baz")
+    conn = mock.Rados.return_value
+    ioctx = conn.open_ioctx.return_value
+
+    rr.close()
+    rr.close()
+    rr.close()
+
+    assert ioctx.close.call_count == 1
+    assert conn.shutdown.call_count == 1


sambacc/ceph/rados.py
```diff
@@ -188,6 +188,7 @@ def close(self) -> None:
             self._closed = True
         if self._connected:
             self._conn.shutdown()
+            self._connected = False
```
anoopcs9, line 191: Good find.

anoopcs9 (dismissed):
lgtm, thanks.

phlogistonjohn:
Just a quick note before I take a look (soon), because the PR description is verbose and sometimes verbosity is a sign of LLMs these days... if the patch was generated (even partially) with LLM based coding tools please document that in the commit with an 'Assisted-by' git trailer (this is similar to what the Linux kernel folks are asking for lately). 
If this is not the case, feel free to disregard my comment.

tests/test_rados_leak.py
```diff
@@ -0,0 +1,114 @@
...
+# is no real cluster; the fake Rados object simply counts connect/shutdown so
+# we can assert that connections are released and do not accumulate.
+
+
+def _fake_cluster_meta(monkeypatch):
+    """Wire up a ClusterMetaRADOSObject backed by a MagicMock rados module.
+
+    Returns (cmo, conn, ioctx) where ``conn`` is the (shared) fake Rados
+    connection and ``ioctx`` its io context, so tests can inspect call counts.
+    """
```
phlogistonjohn: Why not make this a fixture?

tests/test_rados_leak.py
```diff
@@ -0,0 +1,114 @@
...
+    with pytest.raises(RuntimeError):
+        with cmo.open(locked=True):
+            raise RuntimeError("boom")
+
+    assert ioctx.unlock.called
+    assert conn.shutdown.call_count == 1
+    assert ioctx.close.call_count == 1
+
+
+def test_rados_object_ref_close_is_idempotent(monkeypatch):
```
phlogistonjohn: I don't see the `monkeypatch` fixture getting used in this function, remove it if it is not used.

phlogistonjohn (changes requested):
Code largely looks OK. I have one suggestion (fixture) and one must fix (cleanup unused fixture) in the test code.
Thanks for looking into this.

dwburks:
Thanks for the review. Yes, this was AI-assisted (Claude). I've added an Assisted-by trailer as you suggested.

I was rebooting the smb container every few days to keep it from OOMing the host, and couldn't find a documented fix or any open issue from someone else running into the same problem. I verified the fix on real hardware before submitting.

I've also addressed the test feedback: removed the unused monkeypatch parameter and converted _fake_cluster_meta into a fixture. Repushed.

sambacc/ceph/rados.py
```diff
@@ -354,8 +355,18 @@ def __enter__(self) -> Self:
...
+            if self._locked:
+                self._rados_obj._unlock(self._lock_name, self._cookie)
+        finally:
+            # Always release the underlying RADOS connection. open() builds a
+            # fresh, fully-connected RADOSObjectRef for every context and no
+            # caller retains it past the `with` block, so failing to close
+            # here leaks one librados client (and its ~dozen threads) per
+            # invocation -- e.g. once per poll in
+            # ctdb.monitor_cluster_meta_changes(). Runs even if _unlock (or
+            # the with-body) raised; close() is idempotent.
```
anoopcs9, line 368: Now that I'm looking at this again, I wonder if we really need such an elaborate inline comment, since the commit message already covers the rationale in detail (a one-liner might suffice). But I don't mind if others don't feel the same.

anoopcs9 (approved):

sambacc/ceph/rados.py
phlogistonjohn: It's probably fine IMO, but the length of that was one of the things that make me think LLMs were used :-)

phlogistonjohn (approved):
lgtm

PR #212, opened 2026-07-20
config: Sync users and groups to passdb on config change

anoopcs9:
_InstanceConfig.same()_ compared shares, globals, and keybridge but not users/groups, so `update-config --watch` never detected changes to user/group data. Additionally, _\_update_config()_ had no code path to re-import users into _/etc/passwd_ or samba's passdb even if a change were detected. This meant that runtime changes to users or groups in the configuration source were silently ignored by the config watcher - requiring a full container restart to take effect.

This PR refactors _InstanceConfig.same()_ by introducing a `DifferenceFlag` enum and a new _compare()_ method that returns a set of flags indicating which aspects of the configuration differ, rather than a simple boolean. It also extracts user sync logic into reusable _sync_sys_users()_ and _sync_passdb_users()_ functions callable without a CLI context. Finally, it adds users/groups comparison to _compare()_ and calls the sync functions from _\_update_config()_ when user/group data changes between iterations.

### DIFF
--- a/sambacc/commands/config.py
+++ b/sambacc/commands/config.py
@@ -109,7 +109,8 @@ def _update_config(
     boolean indicating if the instance configs differed.
     """
     # has the config changed?
-    changed = not current.same(previous, log_diff=_log_diff)
+    differences = current.compare(previous, log_diff=_log_diff)
+    changed = bool(differences)
     # ensure share paths exist
     if changed and ensure_paths:
         for share in current.shares():
--- a/sambacc/commands/config.py
+++ b/sambacc/commands/config.py
@@ -126,6 +126,13 @@ def _update_config(
         _logger.info("Updating samba configuration")
         loader = nc.NetCmdLoader()
         loader.import_config(current)
+    # update users and groups if they changed
+    if config.DifferenceFlag.USERS_GROUPS in differences:
+        _logger.info("Updating users and groups")
+        from .users import sync_sys_users, sync_passdb_users
+
+        sync_sys_users(current)
+        sync_passdb_users(current)
     # notify smbd of changes
     if changed and notify_server:
         try:
--- a/sambacc/commands/users.py
+++ b/sambacc/commands/users.py
@@ -18,8 +18,36 @@
 
 import sambacc.passdb_loader as passdb
 import sambacc.passwd_loader as ugl
+from sambacc import config
 
 from .cli import commands, setup_steps, Context
+
+
+def sync_sys_users(
+    iconfig: config.InstanceConfig,
+    passwd_path: str = "/etc/passwd",
+    group_path: str = "/etc/group",
+) -> None:
+    """Import users and groups from an InstanceConfig to the passwd and
+    group files.
+    """
+    etc_passwd_loader = ugl.PasswdFileLoader(passwd_path)
+    etc_group_loader = ugl.GroupFileLoader(group_path)
+    etc_passwd_loader.read()
+    etc_group_loader.read()
+    for u in iconfig.users():
+        etc_passwd_loader.add_user(u)
+    for g in iconfig.groups():
+        etc_group_loader.add_group(g)
+    etc_passwd_loader.write()
+    etc_group_loader.write()
+
+
+def sync_passdb_users(iconfig: config.InstanceConfig) -> None:
+    """Import users into samba's passdb."""
+    smb_passdb_loader = passdb.PassDBLoader()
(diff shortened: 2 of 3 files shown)


sambacc/commands/config.py
```diff
@@ -44,6 +44,19 @@
 _logger = logging.getLogger(__name__)
 
 
+def _users_groups_changed(
+    current: config.InstanceConfig,
+    previous: typing.Optional[config.InstanceConfig],
+) -> bool:
+    if previous is None:
+        return False
+    from sambacc.config import _users_groups_data
```
phlogistonjohn: We shouldn't import "private" functions, even if they're not at the top scope. Please consider renaming `_user_groups_data` to `user_groups_data` if you need it in other files.

sambacc/commands/config.py
```diff
@@ -44,6 +44,19 @@
...
+
+    return _users_groups_data(current.gconfig) != _users_groups_data(
+        previous.gconfig
+    )
```
phlogistonjohn: Why not use InstanceConfig.same here now that you have updated it to include users and groups?

phlogistonjohn:
Thinking about it a little more, let's refactor InstanceConfig.same into a function that returns a set of difference flags:

```python
# UNTESTED

class DifferenceFlag(enum.Enum):
    TYPE = 1
    INSTANCE = 2
    SHARES = 3
    GLOBALS = 4
    USERS_GROUPS = 5
    KEYBRDIGE = 6
    # ....

# -------------------

    def compare(
        self,
        other: typing.Any,
        *,
        log_diff: typing.Optional[LogDiffFunc] = None,
    ) -> set[DifferenceFlag]:
        if not log_diff:
            log_diff = _no_op_log_diff
        if not isinstance(other, InstanceConfig):
            log_diff("not an InstanceConfig", self, other)
            return set([Different.TYPE])
        differences = set()
        if self.iconfig != other.iconfig:
            log_diff(
                "instance config values differ", self.iconfig, other.iconfig
            )
            differences.add(DifferenceFlag.INSTANCE)
        self_shares = _shares_data(self.gconfig, self.iconfig)
        other_shares = _shares_data(other.gconfig, other.iconfig)
        if self_shares != other_shares:
            log_diff("shares values differ", self_shares, other_shares)
            differences.add(DifferenceFlag.SHARES)
        self_globals = _globals_data(self.gconfig, self.iconfig)
        other_globals = _globals_data(other.gconfig, other.iconfig)
        if self_globals != other_globals:
            log_diff("globals values differ", self_globals, other_globals)
            differences.add(DifferenceFlag.GLOABALS)
        # keybridge values are be held outside of the iconfig/gconfig
        self_kbridge = _kbridge_data(self.keybridge_config())
        other_kbridge = _kbridge_data(other.keybridge_config())
        if self_kbridge != other_kbridge:
            log_diff("keybridge values differ", self_kbridge, other_kbridge)
            differences.add(DifferenceFlag.KEYBRIDGE)
        self_ug = _users_groups_data(self.gconfig)
        other_ug = _users_groups_data(other.gconfig)
        if self_ug != other_ug:
            log_diff("users/groups values differ", self_ug, other_ug)
            differences.add(DifferenceFlag.USERS_GROUPS)
        return differences
   
    def same(
        self,
        other: typing.Any,
        *,
        log_diff: typing.Optional[LogDiffFunc] = None,
    ) -> bool:
        return not self.compare(other, log_diff=log_diff)
        
```

Then we can do 1 comparison and either reuse or cache the difference flags so other code doesn't need to re-compare things.  WDYT?

sambacc/commands/config.py
anoopcs9: Alright, the refactored version defines it directly here in this file.

sambacc/commands/config.py
anoopcs9: With the updated version, we no longer have this function at all.

anoopcs9 (commented):
> Thinking about it a little more, let's refactor InstanceConfig.same into a function that returns a set of difference flags: 
> [ . . . ]
> Then we can do 1 comparison and either reuse or cache the difference flags so other code doesn't need to re-compare things. WDYT?

Nice thinking, I've updated following the pseudo code shared above.

phlogistonjohn (approved):
lgtm, thanks

anoopcs9:
@Mergifyio rebase

avanthakkar (approved):

PR #215, opened 2026-07-22
support split passwd & groups files

phlogistonjohn:
This began purely as a set of experiments on sambacc to see if we could solve the issue that the configwatch could not update the /etc/passwd file of the smbd container on ceph. I began looking at the altfiles nsswitch module (available as a package on our supported container distros) and because I tend to think in code it was easier to try things and eventually got to the point where I might as well make a PR.

TODO
* [x] Clean up patches into something reviewable
* [x] Investigate / test when running in CTDB with configwatch (leader / not leader, etc)
* [x] Make a decision WRT the mutable passwd/group files needing to exist in order to write them - either fix <s>or decide its an OK prereq</s>
* [x] test split up config watch functions

What I manually tested:
* Build samba-container with the `nss-altfiles` package included
* A user-auth single node "cluster" running on ceph (with a few hacks applied to cephadm)
 * Basic setup worked (after fixes)
 * User added to the list worked (after fixes)
 * ctdb enabled configured for local users/gropus - deploy and update

### DIFF
--- a/sambacc/commands/config.py
+++ b/sambacc/commands/config.py
@@ -134,8 +134,8 @@ def _update_config(
 
         sync_sys_users(
             current,
-            ctx.cli.etc_passwd_path,
-            ctx.cli.etc_group_path,
+            ctx.cli.passwd_location,
+            ctx.cli.group_location,
         )
         sync_passdb_users(current)
     # notify smbd of changes
--- a/sambacc/commands/config.py
+++ b/sambacc/commands/config.py
@@ -98,6 +98,7 @@ UpdateFunc = typing.Callable[..., UpdateResult]
 
 
 def _update_config(
+    ctx: Context,
     current: config.InstanceConfig,
     previous: typing.Optional[config.InstanceConfig],
     ensure_paths: bool = True,
@@ -131,7 +132,11 @@ def _update_config(
         _logger.info("Updating users and groups")
         from .users import sync_sys_users, sync_passdb_users
 
-        sync_sys_users(current)
+        sync_sys_users(
+            current,
+            ctx.cli.etc_passwd_path,
+            ctx.cli.etc_group_path,
+        )
         sync_passdb_users(current)
     # notify smbd of changes
     if changed and notify_server:
@@ -241,13 +246,14 @@ class Trigger:
 @commands.command(name="update-config", arg_func=_update_config_args)
 def update_config(ctx: Context) -> None:
     _get_config = functools.partial(_read_config, ctx)
+    _uconfig = functools.partial(_update_config, ctx)
     trigger = Trigger()
 
     if ctx.instance_config.with_ctdb:
         _logger.info("enabling ctdb support: will check for leadership")
-        trigger.add("samba", _exec_if_leader(ctx, _update_config))
+        trigger.add("samba", _exec_if_leader(ctx, _uconfig))
     else:
-        trigger.add("samba", _update_config)
+        trigger.add("samba", _uconfig)
     if pids_dir := ctx.cli.signal_pids_dir:
         trigger.add("pids", functools.partial(_signal_pids_dir, pids_dir))
 
--- a/sambacc/commands/config.py
+++ b/sambacc/commands/config.py
@@ -17,6 +17,7 @@
 #
 
 import argparse
+import dataclasses
 import functools
 import logging
 import os
@@ -26,9 +27,11 @@ import subprocess
 import sys
 import typing
 
+
 from sambacc import config
 from sambacc import samba_cmds
 from sambacc.simple_waiter import watch
+from sambacc.typelets import Self
 import sambacc.netcmd_loader as nc
 import sambacc.paths as paths
 
@@ -91,6 +94,56 @@ def _read_config(ctx: Context) -> config.InstanceConfig:
         require_validation=ctx.require_validation,
         opener=ctx.opener,
     ).get(ctx.cli.identity)
+
+
+@dataclasses.dataclass(frozen=True)
+class ChangeContext:
+    current: config.InstanceConfig
+    previous: typing.Optional[config.InstanceConfig] = None
+    differences: typing.Optional[set[config.DifferenceFlag]] = None
+    applied: int = 0
+
+    @property
+    def changed(self) -> bool:
+        return bool(self.differences)
+
+    @property
+    def updated(self) -> bool:
+        return bool(self.applied)
+
+    def _all_differences(self) -> set[config.DifferenceFlag]:
+        return set() if self.differences is None else self.differences
+
+    def match(
+        self,
+        flags: typing.Union[
+            config.DifferenceFlag, typing.Iterable[config.DifferenceFlag]
+        ],
+        *,
+        wildcard_type_diff: bool = True,
+    ) -> bool:
+        """Return true if any of the flags specified are in the differences
+        set. If wildcard_type_diff is true, it treats a type mismatch (e.g.
+        previous is None) as a flag match.
+        """
+        adiff = self._all_differences()
+        if wildcard_type_diff and config.DifferenceFlag.TYPE in adiff:
+            return True
+        if isinstance(flags, config.DifferenceFlag):
+            flags = [flags]
+        return any(f in adiff for f in flags)
+
+    def diff(self) -> Self:
+        # cached differences
+        if self.differences is not None:
+            return self
+        # no differences set yet, get fresh
+        differences = self.current.compare(self.previous, log_diff=_log_diff)
+        _logger.debug("config differences found: %r", differences)
+        return dataclasses.replace(self, differences=differences)
+
+    def set_applied(self) -> Self:
+        return dataclasses.replace(self, applied=self.applied + 1)
 
 
 UpdateResult = typing.Tuple[typing.Optional[config.InstanceConfig], bool]
--- a/sambacc/commands/config.py
+++ b/sambacc/commands/config.py
@@ -214,20 +214,29 @@ def _samba_config(chctx: ChangeContext) -> ChangeContext:
 
 
 def _ctr_users_groups(ctx: Context, chctx: ChangeContext) -> ChangeContext:
-    """Update host (container) defined users and groups if the users/groups
-    have changed.
+    """Update host (container) defined users and groups (passwd file) if the
+    users/groups have changed.
     """
     if not chctx.match(config.DifferenceFlag.USERS_GROUPS):
         return chctx
 
     # update users and groups if they changed
     _logger.info("Updating users and groups")
-
     sync_sys_users(
         chctx.current,
         ctx.cli.passwd_location,
         ctx.cli.group_location,
     )
+    return chctx.set_applied()
+
+
+def _sync_passdb(ctx: Context, chctx: ChangeContext) -> ChangeContext:
+    """Update samba's passdb if the users/groups have changed."""
+    if not chctx.match(config.DifferenceFlag.USERS_GROUPS):
+        return chctx
+
+    # update users and groups if they changed
+    _logger.info("Updating samba passdb")
     sync_passdb_users(chctx.current)
     return chctx.set_applied()
 
@@ -317,6 +326,7 @@ def update_config(ctx: Context) -> None:
     )
     cb_samba_config = _samba_config
     cb_ctr_users_groups = functools.partial(_ctr_users_groups, ctx)
+    cb_passdb = functools.partial(_sync_passdb, ctx)
     trigger = Trigger()
 
     if ctx.instance_config.with_ctdb:
@@ -324,10 +334,12 @@ def update_config(ctx: Context) -> None:
         cb_share_paths_shared = functools.partial(
             _when_leader, cb_share_paths_shared
         )
+        cb_passdb = functools.partial(_when_leader, cb_passdb)
         cb_samba_config = functools.partial(_when_leader, cb_samba_config)
     trigger.add("paths_shared", cb_share_paths_shared)
     trigger.add("paths_local", cb_share_paths_local)
     trigger.add("users_groups", cb_ctr_users_groups)
+    trigger.add("sync_passdb", cb_passdb)
     trigger.add("samba", cb_samba_config)
     if pids_dir := ctx.cli.signal_pids_dir:
         trigger.add("pids", functools.partial(_signal_pids_dir, pids_dir))
--- a/sambacc/commands/config.py
+++ b/sambacc/commands/config.py
@@ -43,6 +43,7 @@ from .cli import (
     perms_handler,
     setup_steps,
 )
+from .users import sync_sys_users, sync_passdb_users
 
 _logger = logging.getLogger(__name__)
 
@@ -96,6 +97,14 @@ def _read_config(ctx: Context) -> config.InstanceConfig:
     ).get(ctx.cli.identity)
 
 
+def _readpid(path: pathlib.Path) -> int:
+    try:
+        return int(path.read_text())
+    except Exception as err:
+        _logger.warning("Unable to read pid file %r: %s", path, err)
+        return -1
+
+
 @dataclasses.dataclass(frozen=True)
 class ChangeContext:
     current: config.InstanceConfig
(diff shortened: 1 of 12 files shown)


sambacc/nsswitch_loader.py
```diff
@@ -62,3 +62,32 @@ def ensure_winbind_enabled(self) -> None:
...
             self.lines[gidx] = "group:    files winbind\n"
+
+    def altfiles_enabled(self) -> bool:
+        pline = self.lines[self.idx["passwd"]]
+        gline = self.lines[self.idx["group"]]
+        return ("altfiles" in pline) and ("altfiles" in gline)
+
+    def ensure_altfiles_enabled(self) -> None:
+        pidx = self.idx["passwd"]
+        if "winbind" not in self.lines[pidx]:
```
anoopcs9:
```suggestion
        if "altfiles" not in self.lines[pidx]:
```

sambacc/nsswitch_loader.py
```diff
@@ -62,3 +62,32 @@ def ensure_winbind_enabled(self) -> None:
...
+            self.lines[pidx] = "passwd:    files altfiles\n"
+        gidx = self.idx["group"]
+        if "winbind" not in self.lines[gidx]:
```
anoopcs9:
```suggestion
        if "altfiles" not in self.lines[gidx]:
```

sambacc/commands/users.py
```diff
@@ -16,23 +16,36 @@
...
+    passwd_location: AltLocation,
+    group_location: AltLocation,
+    nsswitch_location: typing.Optional[AltLocation] = None,
 ) -> None:
     """Import users and groups from an InstanceConfig to the passwd and
     group files.
     """
-    etc_passwd_loader = ugl.PasswdFileLoader(passwd_path)
-    etc_group_loader = ugl.GroupFileLoader(group_path)
+    if passwd_location.altfiles or passwd_location.altfiles:
```
anoopcs9:
```suggestion
    if passwd_location.altfiles or group_location.altfiles:
```

sambacc/commands/users.py
```diff
@@ -42,6 +55,11 @@ def sync_sys_users(
     etc_passwd_loader.write()
     etc_group_loader.write()
 
+    if passwd_location.link_path:
+        os.symlink(passwd_location.writable, passwd_location.link_path)
+    if group_location.link_path:
+        os.symlink(group_location.writable, group_location.link_path)
```
anoopcs9: The symlink can exist from the second iteration onwards, right?

anoopcs9 (changes requested):

sambacc/commands/users.py
```diff
@@ -16,23 +16,36 @@
...
+    nsswitch_location: typing.Optional[AltLocation] = None,
 ) -> None:
     """Import users and groups from an InstanceConfig to the passwd and
     group files.
     """
-    etc_passwd_loader = ugl.PasswdFileLoader(passwd_path)
-    etc_group_loader = ugl.GroupFileLoader(group_path)
+    if passwd_location.altfiles or passwd_location.altfiles:
+        if not nsswitch_location:
+            nsswitch_location = AltLocation(nsswitch.NSSWITCH_DEST)
```
anoopcs9: What are we doing here with this check?

sambacc/commands/users.py
```diff
@@ -16,23 +16,36 @@
...
+        nss = nsswitch.find()
+        if not nss.altfiles_enabled():
+            nss.ensure_altfiles_enabled()
+            nss.write(nsswitch_location.writable)
```
anoopcs9: I think this can be a one-time task inside _\_import_nsswitch()_ with an additional early check inside _\_auto_choose_nsswitch()_.

sambacc/commands/users.py
phlogistonjohn: Fixed

sambacc/nsswitch_loader.py
phlogistonjohn: Fixed

sambacc/commands/users.py
phlogistonjohn: I wanted to be consistent with the types passed in to this function and use AltLocation, but not require an nsswitch_location argument. It defaults to None. So if it's none we create a default AltLocation with the typical location for writing the nnswitch.conf file.

sambacc/commands/users.py
phlogistonjohn: Generally I want sambacc to be as "repeatable" as possible. Reading and checking the nsswitch file after it's been updated is a little redundant but I like the repeatability it gives.

anoopcs9: But we don't do it for _winbind_ checks. For consistency and efficiency, I would like to stand by my suggestion.

sambacc/commands/users.py
anoopcs9: I’m not asking why it's coded this way; rather, why does _nsswitch_location_ need to exist in the first place for this function itself?

phlogistonjohn: I wanted the future option to pass it in. It's not used now but I'd like to move away from "locked in" paths for unit testing purposes.

sambacc/commands/users.py
phlogistonjohn: OK, done. But I'm feeling far to lazy to fold this back into the earlier patches.

sambacc/commands/users.py
phlogistonjohn: irrelevant now given the other change

sambacc/commands/users.py
phlogistonjohn: resolved, thanks

sambacc/nsswitch_loader.py
phlogistonjohn: resolved

phlogistonjohn:
Did you have an opinion about having to create (empty) passwd/group files outside of sambacc?  I am leaning towards making sambacc handle it but obviously not having that extra handling keeps the code simple.

But before I get back to that topic code-wise I want to look into the ctdb behavior as that's more subtle.

phlogistonjohn:
After looking over the code and doing a practical test run I confirmed that when ctdb is enabled the configwatch only updates the passwd files on the ctdb leader node.

Also on my hacky ceph patch it doesn't update the nsswitch.conf, but that a bit outside the scope of these patches.

anoopcs9:
> Did you have an opinion about having to create (empty) passwd/group files outside of sambacc? I am leaning towards making sambacc handle it but obviously not having that extra handling keeps the code simple.

I agree, let sambacc ensure the files are present.

anoopcs9:
> After looking over the code and doing a practical test run I confirmed that when ctdb is enabled the configwatch only updates the passwd files on the ctdb leader node.

OK, then I think it's just a matter of extracting _sync_sys_users()_ from __update_config()_ into a separate, unconditional trigger (with a similar detection of `DifferenceFlag`) inside _update_config()_, as the first one.

phlogistonjohn:
Sometimes your changes fix bugs.
Sometimes your changes have bugs.
Sometimes your bugs hide bugs.
Sometimes your changes have bugs that have bugs that hid bugs...  :sob:

phlogistonjohn:
it finally passed my local tests with ctdb enabled. of course now it's a total mess that reflects my mistakes and stumbling blocks along the way. all that needs to be cleaned up. But I can finally deploy on ceph (see https://github.com/ceph/ceph/compare/main...phlogistonjohn:ceph:jjm-EXP-etc-altpaths ) with ctdb and have it update new users live :sweat:

phlogistonjohn:
Having passed my local testing I'm marking this as ready for review.

I'm currently building a branch in ceph that consumes a custom samba-container build with this branch of sambacc with the goal of doing a teuthology run to regression test these specific changes. IMO, teuthology is the most comprehensive deployment test suite right now.

sambacc/commands/users.py
```diff
@@ -19,29 +19,51 @@
...
+        # if we are not using alftiles we expect an existing readable passwd
+        # file. re-raise the exception if we are not using altfiles
+        if not passwd_location.altfiles:
+            raise
+    try:
+        etc_group_loader.read()
+    except FileNotFoundError:
+        # if we are not using alftiles we expect an existing readable group
+        # file. re-raise the exception if we are not using altfiles
+        if not passwd_location.altfiles:
```
anoopcs9:
```suggestion
        if not group_location.altfiles:
```

sambacc/commands/users.py
```diff
@@ -19,29 +19,51 @@
...
+        etc_passwd_loader.read()
+    except FileNotFoundError:
+        # if we are not using alftiles we expect an existing readable passwd
+        # file. re-raise the exception if we are not using altfiles
+        if not passwd_location.altfiles:
+            raise
+    try:
+        etc_group_loader.read()
+    except FileNotFoundError:
+        # if we are not using alftiles we expect an existing readable group
```
anoopcs9:
```suggestion
        # if we are not using altfiles we expect an existing readable group
```

sambacc/config.py
```diff
@@ -664,6 +690,29 @@ def options(self) -> dict[str, str]:
         return {k: v for k, v in self._pconf.items() if k not in filter_keys}
 
 
+class ShareOrigin(enum.Enum):
+    UNKOWN = "unknown"
```
anoopcs9:
```suggestion
    UNKNOWN = "unknown"
```

sambacc/config.py
```diff
@@ -514,6 +522,24 @@ def permissions_config(self) -> PermissionsConfig:
...
         return PermissionsConfig({})
 
+    def meta(self) -> ShareMeta:
+        """Return a share meta configuration for the share."""
+        # each share can have it's own meta config,
+        # but if it does not it will default to the instance's
+        # share meta config
+        try:
+            share_perms = self._cfg()["meta"]
+            return ShareMeta(share_perms)
```
anoopcs9:
```suggestion
            share_meta = self._cfg()["meta"]
            return ShareMeta(share_meta)
```

sambacc/config.py
```diff
@@ -514,6 +522,24 @@ def permissions_config(self) -> PermissionsConfig:
...
+        except KeyError:
+            pass
+        try:
+            instance_perms = self.iconfig["sharemeta"]
+            return ShareMeta(instance_perms)
```
anoopcs9:
```suggestion
            instance_meta = self.iconfig["sharemeta"]
            return ShareMeta(instance_meta)
```

sambacc/commands/config.py
```diff
@@ -213,58 +288,89 @@ def _signal_pids_dir(
...
+    cbname = getattr(cb, "__name__", f"?{cb}")
+    if cbname.startswith("?") and hasattr(cb, "func"):
+        cbname = getattr(cb.func, "__name__", f"?{cb}")
+    with best_leader_locator(chctx.current) as ll:
+        if not ll.is_leader():
+            _logger.info("skipping %s. node not leader", cbname)
+            return chctx
+        _logger.info("executing %s. node is leader", cbname)
+        return cb(chctx)
+    return chctx
```
anoopcs9, line 287: This is never reachable.

sambacc/commands/config.py
```diff
@@ -93,104 +97,175 @@ def _read_config(ctx: Context) -> config.InstanceConfig:
...
+        if not path:
+            continue
+        _logger.info(f"Ensuring share path: {path}")
+        paths.ensure_share_dirs(path)
+        _logger.info(f"Updating permissions if needed: {path}")
+        perms_handler(share.permissions_config(), path).update()
+
+
+def _samba_config(
+    ctx: Context, chctx: ChangeContext, ensure_paths: bool = True
```
anoopcs9: _ctx_ and _ensure_paths_ seems to be unused.

sambacc/commands/config.py
```diff
@@ -213,58 +288,89 @@ def _signal_pids_dir(
             os.kill(pid, signal.SIGHUP)
         except OSError as err:
             _logger.warning("Failed to SIGHUP %d: %s", pid, err)
-    return current, changed
+    return chctx.set_applied()
+
+
+def _when_leader(cb: ChangeCheck, chctx: ChangeContext) -> ChangeContext:
+    # NOTE: it's important to mainatin leader status while running the callback
```
anoopcs9:
```suggestion
    # NOTE: it's important to maintain leader status while running the callback
```

sambacc/paths.py
```diff
@@ -55,3 +55,24 @@ def ensure_share_dirs(path: str, root: str = "/") -> None:
...
+        return
+    except FileExistsError:
+        # dst exists but we will ignore it if update is False
+        if not update:
+            return
+    lnk = os.readlink(dst)  # this will raise if dst is not a link
+    if lnk == src:
+        return
+    try:
+        os.unlink(dst)
```
anoopcs9: The subsequent _os.symlink()_ fails with `FileExistsError` if _os.unlink()_ fails, hiding the real error.

sambacc/config.py
```diff
@@ -664,6 +690,29 @@ def options(self) -> dict[str, str]:
...
+    @property
+    def origin(self) -> ShareOrigin:
+        """Return the origin (kind of path) that a share is based on.
+        Defaults to 'shared' as historically that's what other parts of sambacc
+        assumed.
+        """
+        try:
+            return ShareOrigin(self._mconf.get("origin", "shared"))
+        except (KeyError, ValueError):
+            return ShareOrigin.UNKOWN
```
anoopcs9:
```suggestion
            return ShareOrigin.UNKNOWN
```

sambacc/commands/config.py
```diff
@@ -93,104 +97,175 @@ def _read_config(ctx: Context) -> config.InstanceConfig:
...
+def _share_paths_shared(chctx: ChangeContext) -> ChangeContext:
+    """Iterate over shares ensuring the paths and permissions are set up.
+    The "shared shares" are shares on shared media (a clustered fs) mounted
+    into a namespace available to this process.  When executed in a clustered
+    samba setup this should only be run by the leader.
+    """
+    if not chctx.match(config.DifferenceFlag.SHARES):
+        return chctx
+    _ensure_share_paths(chctx.shared_shares())
+    return chctx.set_applied()
```
anoopcs9: _share_paths_local()_ and _share_paths_shared()_ are nearly identical. Why can't we have a common _share_paths()_ accepting _config.ShareOrigin_? This way we could also remove _local_shares()_ and _shared_shares()_ moving that logic directly within _share_paths()_.

sambacc/commands/config.py
```diff
@@ -213,58 +288,89 @@ def _signal_pids_dir(
...
+
+def _when_leader(cb: ChangeCheck, chctx: ChangeContext) -> ChangeContext:
+    # NOTE: it's important to mainatin leader status while running the callback
+    # thus the context manager.
+    # CTDB enablement and the need for leader detection should not change
+    # while the process is running. Thus, _when_leader can wrap all future uses
+    # of a change check.
+    cbname = getattr(cb, "__name__", f"?{cb}")
+    if cbname.startswith("?") and hasattr(cb, "func"):
+        cbname = getattr(cb.func, "__name__", f"?{cb}")
```
anoopcs9: All this for printing the name in logs? I would suggest passing the name instead from the registration side.

anoopcs9 (changes requested):

sambacc/commands/config.py
phlogistonjohn: it is reachable because context managers have the capability to handle exceptions. if cb raised and the __exit__ method of the context manager suppresses the exception we will jump to the next line after the context manager.

sambacc/config.py
phlogistonjohn: Uhhhhhhg. I always do that. :-(

sambacc/paths.py
phlogistonjohn: Hiding the real error? That was intentional because if unlink fails I don't find it's error code all that interesting but I do find the failure of the 2nd symlink() call interesting. Does that make sense?

sambacc/commands/config.py
phlogistonjohn: I'm confused. 
_share_paths_local must be run on all nodes
_shares_paths_shared must only be run on the leader (when ctdb is enabled)
The bulk of the share creation logic is shared w/in _ensure_share_paths.

sambacc/commands/config.py
phlogistonjohn: fair

sambacc/commands/config.py
anoopcs9: Ok, I missed that part.

sambacc/commands/config.py
anoopcs9: I don't think my suggestion is affecting the leader-only behaviour. Let me paste the diff I have in my mind:
```diff --git a/sambacc/commands/config.py b/sambacc/commands/config.py
index ca4b878..64e79cc 100644
--- a/sambacc/commands/config.py
+++ b/sambacc/commands/config.py
@@ -154,45 +154,23 @@ class ChangeContext:
     def set_applied(self) -> Self:
         return dataclasses.replace(self, applied=self.applied + 1)
 
-    def local_shares(self) -> list[config.ShareConfig]:
-        return [
-            s
-            for s in self.current.shares()
-            if s.meta().origin is config.ShareOrigin.LOCAL
-        ]
-
-    def shared_shares(self) -> list[config.ShareConfig]:
-        return [
-            s
-            for s in self.current.shares()
-            if s.meta().origin is config.ShareOrigin.SHARED
-        ]
-
-
 ChangeCheck = typing.Callable[[ChangeContext], ChangeContext]
 
 
-def _share_paths_local(chctx: ChangeContext) -> ChangeContext:
-    """Iterate over shares ensuring the paths and permissions are set up.
-    The "local shares" are shares on local media, mounted into a namespace
-    available to this process.  When executed in a clustered samba setup this
-    should be executed by all nodes in the cluster.
-    """
-    if not chctx.match(config.DifferenceFlag.SHARES):
-        return chctx
-    _ensure_share_paths(chctx.local_shares())
-    return chctx.set_applied()
-
-
-def _share_paths_shared(chctx: ChangeContext) -> ChangeContext:
-    """Iterate over shares ensuring the paths and permissions are set up.
-    The "shared shares" are shares on shared media (a clustered fs) mounted
-    into a namespace available to this process.  When executed in a clustered
-    samba setup this should only be run by the leader.
+def _share_paths(
+    chctx: ChangeContext, origin: config.ShareOrigin
+) -> ChangeContext:
+    """Ensure share paths and permissions exist for shares of the
+    given origin type.
     """
     if not chctx.match(config.DifferenceFlag.SHARES):
         return chctx
-    _ensure_share_paths(chctx.shared_shares())
+    shares = [
+        s
+        for s in chctx.current.shares()
+        if s.meta().origin is origin
+    ]
+    _ensure_share_paths(shares)
     return chctx.set_applied()
 
 
@@ -340,8 +318,12 @@ class Trigger:
 @commands.command(name="update-config", arg_func=_update_config_args)
 def update_config(ctx: Context) -> None:
     _get_config = functools.partial(_read_config, ctx)
-    cb_share_paths_local = _share_paths_local
-    cb_share_paths_shared = _share_paths_shared
+    cb_share_paths_local = functools.partial(
+        _share_paths, origin=config.ShareOrigin.LOCAL
+    )
+    cb_share_paths_shared = functools.partial(
+        _share_paths, origin=config.ShareOrigin.SHARED
+    )
     cb_samba_config = functools.partial(_samba_config, ctx)
     cb_ctr_users_groups = functools.partial(_ctr_users_groups, ctx)
     cb_passdb = functools.partial(_sync_passdb, ctx)
```

sambacc/paths.py
anoopcs9: The issue is that if unlink fails, the symlink call always fails with `FileExistsError` - which doesn't tell you anything new. You already know the file exists. The interesting information (why couldn't we remove it) is lost. Even a debug log of the unlink error would help diagnose permission or filesystem issues without changing the control flow.

sambacc/commands/config.py
phlogistonjohn: I see, thanks.

phlogistonjohn:
https://pulpito-ng.ceph.com/runs/avan-2026-07-29_13:03:07-orch:cephadm:smb-wip-phlogistonjohn-sambacc-test-1-distro-default-trial

Using ceph as a testbed for checking for regressions I'd say there's only 1 unexpected error and I think it's just a flake.

sambacc/commands/users.py
phlogistonjohn: done

sambacc/commands/users.py
phlogistonjohn: done

sambacc/config.py
phlogistonjohn: done

sambacc/config.py
phlogistonjohn: done

sambacc/config.py
phlogistonjohn: done

sambacc/commands/config.py
phlogistonjohn: done

sambacc/commands/config.py
phlogistonjohn: done

sambacc/config.py
phlogistonjohn: done

sambacc/paths.py
phlogistonjohn: :shrug: done

sambacc/commands/config.py
phlogistonjohn: done, but with my preferred ordering to the args so that the partials are consistent

sambacc/commands/config.py
phlogistonjohn: I started doing this and realize it would break the layering because of how we construct partial fns using this as a wrapper. It has no clean access to the other var. I made a small simplification tho.

anoopcs9: OK.

sambacc/commands/config.py
```diff
@@ -93,104 +97,153 @@ def _read_config(ctx: Context) -> config.InstanceConfig:
...
+
+
+def _share_paths(
+    origin: config.ShareOrigin, chctx: ChangeContext
+) -> ChangeContext:
+    """Iterate over shares matching the given origin type, ensuring the paths
+    and permissions are set up.
+    Shares are separated by origin in order to only run this function on
+    the appropriate node(s).
+    For example: Shares in the local origin can, and should, be set up on al
```
anoopcs9:
```suggestion
    For example: Shares in the local origin can, and should, be set up on all
```

anoopcs9 (commented):
A bisectability question: commit `sambacc/commands: make nsswitch setup more consistent` references _ctx.cli.passwd_location_ but that attribute doesn't exist until `sambacc/commands: rename --etc-... command line options` renames the CLI option. Can these be squashed (by relevant hunks) or reordered?

And with that typo fixed, I think this is in a good shape.

sambacc/commands/config.py
phlogistonjohn: Bah. I reorded those patches last night to make it more logical review-wise but forgot about that. I've fixed the ordering so that the code is consistent even if the patch order is suboptimal.

phlogistonjohn:
Related (draft) ceph PR: https://github.com/ceph/ceph/pull/70679

sambacc/commands/config.py
phlogistonjohn: Both issues should be fixed now.

sambacc/commands/config.py
```diff
@@ -93,104 +97,153 @@ def _read_config(ctx: Context) -> config.InstanceConfig:
...
+def _share_paths(
+    origin: config.ShareOrigin, chctx: ChangeContext
+) -> ChangeContext:
+    """Iterate over shares matching the given origin type, ensuring the paths
+    and permissions are set up.
+    Shares are separated by origin in order to only run this function on
+    the appropriate node(s).
+    For example: Shares in the local origin can, and should, be set up on all
+    nodes in a CTDB cluster while shares in the shared origin should only be
+    set up one (by the leader).
```
anoopcs9:
```suggestion
    set up once (by the leader).
```

anoopcs9 (dismissed):
lgtm, thanks.

sambacc/config.py
```diff
@@ -664,6 +690,29 @@ def options(self) -> dict[str, str]:
...
+    @property
+    def origin(self) -> ShareOrigin:
+        """Return the origin (kind of path) that a share is based on.
+        Defaults to 'shared' as historically that's what other parts of sambacc
+        assumed.
+        """
+        try:
+            return ShareOrigin(self._mconf.get("origin", "shared"))
+        except (KeyError, ValueError):
+            return ShareOrigin.UNKNOWN
```
avanthakkar, line 713: Shouldn't we also update `conf-v0.schema.yaml` for the new meta/sharemeta keys?  Since schema validation runs by default, I think any config using meta/sharemeta currently would fail to load with `ValidationError`.  Maybe add a `share_meta_config` def (like `permissions_config`) with an origin enum and reference it from both places?

phlogistonjohn: Good catch, but I think this PR is already too big. Consider ShareOrigin support a prerequsite for the refactoring (splitting) the shares into  leader-only and all-node phases in the update config loop. Since nothing even tries to configure the share meta right now we can finalize the configuration in a follow up PR.

sambacc/commands/config.py
phlogistonjohn: fixed

sambacc/config.py
avanthakkar: Sure. Sounds good to me

anoopcs9 (approved):

sambacc/config.py
phlogistonjohn: https://github.com/samba-in-kubernetes/sambacc/issues/220

avanthakkar (approved):
LGTM

PR #217, opened 2026-07-27
Support adding local users to local groups

### LINKED ISSUE #101: Support to add local users to local groups
See: samba-in-kubernetes/samba-container#163

anoopcs9:
Add support for specifying group membership in local (non-domain) group entries via a new `members` field. This allows configurations like:

  ```json
  "groups": {
      "all_entries": [
          {"name": "readers", "gid": 3000, "members": ["alice", "bob"]},
          {"name": "writers", "gid": 3001, "members": ["alice"]}
      ]
  }
```

Previously, the members field (4th field) in _/etc/group_ lines was always written as an empty string. Groups without members defined keep the previous behavior. _GroupFileLoader.add_group()_ is also updated to modify the members field of existing group lines in place, so that membership changes are picked up during config re-syncs via `update-config`.

fixes #101

### DIFF
--- a/sambacc/config.py
+++ b/sambacc/config.py
@@ -614,9 +614,12 @@ class GroupEntry:
         self.entry_num = num
         self._gid = grec.get("gid")
         self.ou = grec.get("ou")
+        self.members: list[str] = grec.get("members", [])
         if self._gid is not None:
             if not isinstance(self._gid, int):
                 raise ValueError("invalid gid value")
+        if not isinstance(self.members, list):
+            raise ValueError("members should contain a list of user names")
 
     @property
     def gid(self) -> int:
@@ -626,7 +629,7 @@ class GroupEntry:
 
     def group_fields(self) -> GroupEntryTuple:
         # fields: name, passwd, gid, members(comma separated)
-        return (self.groupname, "x", str(self.gid), "")
+        return (self.groupname, "x", str(self.gid), ",".join(self.members))
 
 
 class DomainConfig:
--- a/sambacc/passwd_loader.py
+++ b/sambacc/passwd_loader.py
@@ -82,7 +82,24 @@ class GroupFileLoader(LineFileLoader):
 
     def add_group(self, group_entry: config.GroupEntry) -> None:
         if group_entry.groupname in self._groupnames:
+            self._update_group_members(group_entry)
             return
         line = "{}\n".format(":".join(group_entry.group_fields()))
         self.lines.append(line)
         self._groupnames.add(group_entry.groupname)
+
+    def _update_group_members(self, group_entry: config.GroupEntry) -> None:
+        new_members = group_entry.group_fields()[3]
+        for idx, line in enumerate(self.lines):
+            if ":" not in line:
+                continue
+            fields = line.rstrip("\n").split(":")
+            if len(fields) < 4:
+                continue
+            if fields[0] != group_entry.groupname:
+                continue
+            if fields[3] == new_members:
+                return
+            fields[3] = new_members
+            self.lines[idx] = "{}\n".format(":".join(fields))
+            return
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -247,6 +247,7 @@ The `groups` section supports one key, `all_entries`, which is a list of
 group entries. Each group entry is as follows:
 * `name` - The user's name.
 * `gid` - Optional integer. Specify the exact Unix GID the group should have.
+* `members` - Optional list of strings. User names that are members of the group.
 
 
 ## Domain Settings Section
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

**Comment by phlogistonjohn** [comment]
Without looking at the code yet, I have one concern that jumps out at me from the example. Here we use a members list embedded in the group definition. In the AD DC config the users definition has an argument `members_of` - https://github.com/samba-in-kubernetes/sambacc/blob/master/docs/configuration.md#domain-users-section 

How hard would it be to use the established pattern of `members_of` in this?

**Comment by anoopcs9** [comment]
> Without looking at the code yet, I have one concern that jumps out at me from the example. Here we use a members list embedded in the group definition. In the AD DC config the users definition has an argument `members_of` - https://github.com/samba-in-kubernetes/sambacc/blob/master/docs/configuration.md#domain-users-section
> 
> How hard would it be to use the established pattern of `members_of` in this?

Thanks for pointing out `member_of` - I wasn't aware of that existing pattern in the domain users config.

I went with `members` on the group entry because the local user/group flow writes directly to _/etc/group_, whose format is inherently group centric: each line is "name:passwd:gid:member1,member2,. . .". Having members on the group entry maps 1:1 to that output, which keeps the config to file path simple - _GroupEntry.group_fields()_ produces the line directly.

Looking at it now, in the AD DC case `member_of` works naturally because provisioning is user by user via `samba-tool group addmembers`. The local case doesn't have that intermediary - we are writing file lines per group, so the group needs to know its members at write time.

Switching to `member_of` on local user entries is doable, but it would require _sync_sys_users()_ to first iterate all users to build a group to members mapping, then pass that into the group writing path. It inverts the data flow and adds complexity for what's otherwise a straightforward file write.

My preference would be to keep the current `members` approach since it maps directly to the _/etc/group_ format and avoids the extra indirection.

**Comment by phlogistonjohn** [comment]
Fair enough. At least we have a logical justification that it is "closer to the source material".

**Comment by anoopcs9** [comment]
@Mergifyio update

File: sambacc/passwd_loader.py [hunk 1]
```diff
@@ -82,7 +82,22 @@ def _update_groupnames_cache(self) -> None:
...
     def add_group(self, group_entry: config.GroupEntry) -> None:
         if group_entry.groupname in self._groupnames:
+            self._update_group_members(group_entry)
             return
         line = "{}\n".format(":".join(group_entry.group_fields()))
         self.lines.append(line)
         self._groupnames.add(group_entry.groupname)
+
+    def _update_group_members(self, group_entry: config.GroupEntry) -> None:
+        new_members = group_entry.group_fields()[3]
```
avanthakkar, line 92: Is it possible that the group file has fewer than 4 colon-separated fields?

sambacc/config.py
```diff
@@ -614,9 +614,12 @@ def __init__(self, iconf: InstanceConfig, grec: dict, num: int):
         self.entry_num = num
         self._gid = grec.get("gid")
         self.ou = grec.get("ou")
+        self.members: list[str] = grec.get("members", [])
```
avanthakkar, line 617: Probably add `members` filed to existing `Groups Section` in configuration.md? Similar to `name` & `gid`

sambacc/passwd_loader.py
anoopcs9: _group_fields()_ always returns a 4-tuple `(name, passwd, gid, members)` (see GroupEntryTuple in [sambacc/config.py](https://github.com/samba-in-kubernetes/sambacc/blob/master/sambacc/config.py#L38)), so the index at line 92 is safe.

However, the concern is valid for lines 99 and 101, where _fields[3]_ is accessed on a line read from the existing group file via _split(":")_ - a malformed line with fewer than 4 fields would raise an IndexError. Were you referring to that case instead? If so we could guard against that by skipping such lines.

sambacc/config.py
anoopcs9: Good catch, will add. Something like:
``` 
* members - Optional list of strings. User names that are members of the group.
```

sambacc/passwd_loader.py
avanthakkar: Ohkay wasn't aware. I  was referring to 92 line only but yes what you mentioned about guarding to avoid IndexError sounds good

avanthakkar (approved):
LGTM

phlogistonjohn (approved):
LGTM,

One, non blocking, thought I have is that I don't see validation that the user names in the list are users defined in the spec. I think this means, in theory, you could add other users or misspelled/invalid user names to the list.

I don't think this is a major problem because the author of the sambacc config has ability to do much more dramatic things to the config but UX-wise it might be a good followup enhancement.

PR #225, opened 2026-08-04
config: Validate and deduplicate group members list

### LINKED ISSUE #222: Add name validation to group members list
LGTM,

One, non blocking, thought I have is that I don't see validation that the user names in the list are users defined in the spec. I think this means, in theory, you could add other users or misspelled/invalid user names to the list.

I don't think this is a major problem because the author of the sambacc config has ability to do much more dramatic things to the config but UX-wise it might be a good followup enhancement.

_Originally posted by @phlogistonjohn in https://github.com/samba-in-kubernetes/sambacc/pull/217#pullrequestreview-4829293185_

anoopcs9:
- Silently deduplicate group members list during construction, preserving insertion order
- Validate that user names in a group's members field are defined in the configuration

fixes #222

### DIFF
--- a/sambacc/config.py
+++ b/sambacc/config.py
@@ -614,12 +614,14 @@ class GroupEntry:
         self.entry_num = num
         self._gid = grec.get("gid")
         self.ou = grec.get("ou")
-        self.members: list[str] = grec.get("members", [])
+        members = grec.get("members", [])
+        if not isinstance(members, list):
+            raise ValueError("members should contain a list of user names")
+        # use dict.fromkeys to deduplicate while preserving order
+        self.members: list[str] = list(dict.fromkeys(members))
         if self._gid is not None:
             if not isinstance(self._gid, int):
                 raise ValueError("invalid gid value")
-        if not isinstance(self.members, list):
-            raise ValueError("members should contain a list of user names")
 
     @property
     def gid(self) -> int:
--- a/sambacc/config.py
+++ b/sambacc/config.py
@@ -314,10 +314,21 @@ class InstanceConfig:
             yield UserEntry(self, entry, n)
 
     def groups(self) -> typing.Iterable[GroupEntry]:
-        user_gids = {u.gid: u for u in self.users()}
+        users = list(self.users())
+        user_gids = {}
+        user_names = set()
+        for u in users:
+            user_gids[u.gid] = u
+            user_names.add(u.username)
         all_groups = self.gconfig.data.get("groups", {}).get("all_entries", {})
         for n, entry in enumerate(all_groups):
             ge = GroupEntry(self, entry, n)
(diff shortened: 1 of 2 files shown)


sambacc/config.py
```diff
@@ -614,12 +614,13 @@ def __init__(self, iconf: InstanceConfig, grec: dict, num: int):
         self.entry_num = num
         self._gid = grec.get("gid")
         self.ou = grec.get("ou")
-        self.members: list[str] = grec.get("members", [])
+        members = grec.get("members", [])
+        if not isinstance(members, list):
+            raise ValueError("members should contain a list of user names")
+        self.members: list[str] = list(dict.fromkeys(members))
```
phlogistonjohn: `list(set(members))` please.

sambacc/config.py
```diff
@@ -314,10 +314,18 @@ def users(self) -> typing.Iterable[UserEntry]:
             yield UserEntry(self, entry, n)
 
     def groups(self) -> typing.Iterable[GroupEntry]:
-        user_gids = {u.gid: u for u in self.users()}
+        users = list(self.users())
+        user_gids = {u.gid: u for u in users}
+        user_names = {u.username for u in users}
```
phlogistonjohn: I think it's probably safe to assume we won't have large lists of users, but if we would I would try to avoid double iteration and use a for loop to build both "caches" up.

phlogistonjohn (changes requested):

sambacc/config.py
phlogistonjohn: Now I see you using sets elsewhere in this PR I start to wonder... why this form? Are you trying to use the dict to preserve order?  I'm not sure it matters but if it does for you... just leave a comment explaining why you constructed it this way vs the simpler form I noted previously.

sambacc/config.py
anoopcs9: Done.

sambacc/config.py
anoopcs9: Ordering of group members doesn't matter, so switched to `list(set(members))`.

anoopcs9: Ah..then it affects existing tests because the assertion is in a particular order. What do you suggest? `dict.fromkeys()` or `list(set(members))` with existing tests modified or `sorted(set(members))`.

phlogistonjohn: I have no strong preference either way. I suppose if you twist my arm I would change the code rather than the test because there's  tiny maybe something else assumes the order is maintained. Just explain the `dict.fromkeys` with a comment if you choose that approach.

anoopcs9: Done.

phlogistonjohn (approved):

avanthakkar (approved):

PR #228, opened 2026-08-04
config: retry import_config when ctdb leader not yet elected

### LINKED ISSUE #227: `update-config --watch` misses initial config import when CTDB leader not yet elected
When `update-config --watch` starts on a CTDB cluster, the first iteration detects a config change (previous=None) and attempts to run `import_config`. However if CTDB hasn't elected a leader yet, the iteration is skipped. After CTDB settles, subsequent iterations see no config change (current == previous) so `import_config` never runs, leaving the samba registry stale.

avanthakkar:
When configwatch starts before CTDB elects a leader, the first iteration skips import_config and sets previous=current. Later iterations see no config change and never retry, hence leaving the registry stale.

Return None from Trigger.__call__ when "not a leader" so the watch loop resets previous and retries until a leader can run import_config.

Fixes: https://github.com/samba-in-kubernetes/sambacc/issues/227

### DIFF
--- a/sambacc/commands/config.py
+++ b/sambacc/commands/config.py
@@ -111,6 +111,7 @@ class ChangeContext:
     previous: typing.Optional[config.InstanceConfig] = None
     differences: typing.Optional[set[config.DifferenceFlag]] = None
     applied: int = 0
+    retry: bool = False
 
     @property
     def changed(self) -> bool:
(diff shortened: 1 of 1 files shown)


avanthakkar:
@phlogistonjohn @anoopcs9 Ping for review

PR #229, opened 2026-08-05
sambacc/commands: Add share_paths to default setup steps

### LINKED ISSUE #58: default 'run' setup tasks don't include path ensure func
This means that either you need to list out all the stock setup steps manually (like: `--setup=config --setup=users --setup=users_passdb --setup=nsswitch --setup=share_paths`) or you don't otherwise automatically get path ensurance/permissions update.

Currently, the setup tasks are a bit of a jumble anyway so we may want to clean it up in general too.

anoopcs9:
Ensure share directories and permissions are set up during default container initialization so that standalone deployments not using configwatch also get automatic path setup.

fixes #58

### DIFF
--- a/sambacc/commands/initialize.py
+++ b/sambacc/commands/initialize.py
@@ -122,6 +122,7 @@ def ensure_share_paths(ctx: Context) -> None:
 
 _default_setup_steps = [
     "config",
+    "share_paths",
     "smb_ctdb",
     "nsswitch_auto",
     "users",


sambacc/commands/initialize.py
```diff
@@ -122,6 +122,7 @@ def ensure_share_paths(ctx: Context) -> None:
 
 _default_setup_steps = [
     "config",
+    "share_paths",
```
phlogistonjohn, line 125: Share paths (may) also try to configure permissions. Should it perhaps come after the users?

anoopcs9: Aren't the permissions set by _share_paths_ purely mode-based rather than ownership-based, so they wouldn't depend on users being present?

phlogistonjohn: Without reading the source rn I think that's right, but I was thinking of being more safe than sorry.

anoopcs9: I've confirmed that `share_paths` only does _os.makedirs_ and _os.fchmod_ - no ownership changes at all. Happy to move it after `users` anyway if you'd prefer the defensive ordering.

phlogistonjohn: *sigh* so then those are inconsistent with the update config loop. :angry: not your fault.
It'll be fine for now then.

anoopcs9:
@Mergifyio rebase

phlogistonjohn (dismissed):

avanthakkar (dismissed):

anoopcs9:
CI run did not happen for all jobs and I ended up force pushing which resulted in clearing the previous approvals (I should've retriggered the checks from UI itself, my bad). @avanthakkar @phlogistonjohn Can I have your approvals one more time?

avanthakkar (approved):

phlogistonjohn (approved):

PR #230, opened 2026-08-13
sambacc/commands: Skip virtual shares when ensuring share paths

evan314159:
The share_paths setup step iterated over every share and called os.makedirs on the configured path. For a share whose path contains a samba variable, such as /shares/home/%U, that creates a directory literally named '%U' on every container start.

_share_paths in commands/config.py already separates shares by origin and never runs for the virtual origin. Apply the same rule here, so 'origin': 'virtual' is honoured by both code paths.


phlogistonjohn:
Interesting idea. I had conceptualized the virtual share origin for things like vfs modules that don't use "real paths" in a kernel mounted file system, but this looks like another interesting use case for that flag.

I'll try to do a proper review soon.

sambacc/commands/initialize.py
```diff
@@ -106,11 +107,13 @@ def _ctdb_etc_files(ctx: Context) -> None:
...
 def ensure_share_paths(ctx: Context) -> None:
     """Ensure the paths defined by the configuration exist."""
-    # currently this is completely ignorant of things like vfs
-    # modules that might "virtualize" the share path. It just
+    # other than shares in the virtual origin, this is ignorant of things
+    # like vfs modules that might "virtualize" the share path. It just
     # assumes that the path in the configuration is an absolute
     # path in the file system.
     for share in ctx.instance_config.shares():
+        if share.meta().origin is sambacc_config.ShareOrigin.VIRTUAL:
```
phlogistonjohn: If this is the only thing used from `sambacc_config` it's a bit more succinct to just `from sambacc.config import ShareOrigin` in the imports.

phlogistonjohn (changes requested):
minor change requested

anoopcs9 (approved):
lgtm, thanks.

It's also worth noting that _\_share_paths()_ in _commands/config.py_ takes the opposite approach: it allows `LOCAL` and `SHARED` rather than denying `VIRTUAL`. So they'd behave differently if a new origin type were added. That's fine here, though - for _ensure_share_paths()_, creating an extra directory is harmless, while missing one is not.

@evan314159 Curious - did this come up after `share_paths` was added to the default setup steps? I imagine that change would have been the first time _ensure_share_paths()_ ran automatically for every share. Would love to know if you're testing with a recent build of samba-container - always helpful to know the context behind a fix like this.

evan314159:
I am a new install migrating an existing file server to Kubernetes.  I tried the servercontainers version first, then tried the samba-in-kubernetes image, and noticed that the container created a %U directory for my home share where the directory is set as …/home/%U (user mounts /home and gets their own home directory).

I am presently not using the operator.  I was initially interested in the operator (even before trying servercontainers) but was scared away by the operator README suggesting the operator may be on limited support. Based on further reading now I don't think I can use the operator for my use cases (maybe can take to a discussion in samba-operator if you'd like to know more).  samba-container works well for a non-HA solution.

I’m using :latest.

phlogistonjohn (approved):

PR #231, opened 2026-09-17
config: Support alternate sources for AD admin password

### LINKED ISSUE #80: addc: support alternate method for supplying the admin password
Currently, the only place sambacc sources the admin password from is in the sambacc config file. This means if you want to keep the password more secure you have to put the entire config file in a secret (or equivalent). This is inconvient if you want to keep your sambacc configurations together in one place.

We could have sambacc source the password from another file and/or an environment variable. Off the top of my head I know that both podman and k8s have methods for passing secret data via those means.

This is not an issue for domain member servers because those already take the join user/pass in a separate file from the configuration.

anoopcs9:
Currently the AD domain admin password can only be specified as a plain string in the sambacc config file. Add support for sourcing the password from a file or environment variable via the `admin_password` field:
- String (existing)
  - `"admin_password": "pass"`
- File
  - `"admin_password": {"file": "/path/to/file"}`
- Env
  - `"admin_password": {"env": "ENV_VAR_NAME"}`

fixes #80


phlogistonjohn (commented):
How does this mesh with --password/INSECURE_JOIN_PASSWORD. Should it?

anoopcs9:
> How does this mesh with --password/INSECURE_JOIN_PASSWORD. Should it?

`--password/INSECURE_JOIN_PASSWORD` is for member server joins (`net ads join`) and already accepts value from CLI/env (plus there's even a separate `--join-file` flag, see linked issue #80). `admin_password` is for AD DC provisioning/join (_samba-tool_) and was previously config-only.

PR #232, opened 2026-09-18
Use libsmbconf Python API for config import

### LINKED ISSUE #46: Import configuration with libsmbconf API
Currently, when sambacc imports samba configuration from it's own JSON based fomat it uses the `net conf import` command line. Future versions of Samba will have a python binding for the smbconf library. This library can be used to read a text based config (smb.conf) or read/write registry based configuration. This would allow us to use the library directly rather than generating a text based configuration and shelling out to `net` command to update the registry.

It may also make post-init configuration updates simpler and easier to see exactly what is changing.

anoopcs9:
Replace the `net conf import` subprocess call with direct use of Samba's _libsmbconf_ Python bindings when available, falling back to the CLI tool when the bindings are missing. Adds an _InstanceConfigStore_ adapter to bridge the config model to the smbconf import layer.

fixes #46


phlogistonjohn:
I have a note in my personal todo list that says:
`fix smbconf with ctdb (wrt talloc stackframe)`

It's 5 months old :weary: 

I think there's an issue with initialization when used on a CTDB enabled configuration. I remember discussing it with @gd and saying something like we could probably work around the issue via a change in the C/Python lib, but I don't (or can't find right now) any other written notes.

Test out our prototype changes on a CTDB enabled instance and see if you get crashes related to the talloc stack. If you do the problem still exists in current Samba and we should fix that before taking these improvments on the sambacc side.

anoopcs9:
> I have a note in my personal todo list that says: `fix smbconf with ctdb (wrt talloc stackframe)`
> 
> It's 5 months old 😩
> 
> I think there's an issue with initialization when used on a CTDB enabled configuration. I remember discussing it with @gd and saying something like we could probably work around the issue via a change in the C/Python lib, but I don't (or can't find right now) any other written notes.
> 
> Test out our prototype changes on a CTDB enabled instance and see if you get crashes related to the talloc stack. If you do the problem still exists in current Samba and we should fix that before taking these improvments on the sambacc side.

Reproduced and created [!4724](https://gitlab.com/samba-team/samba/-/merge_requests/4724) as a probable fix.

